PR #60, opened 2014-12-24
expose set_configuration

heavyk:
I have a usb random number generator and it requires that set_configurtion is both called before the config descriptor (obviously). it's a pretty minimal change. (only caveat is if you're gonna configure with a predefined device config, open should be called with false.)

it's mentioned in the readme only initial configurations are supported, ... well here it is if anyone needs it.


kevinmehall:
Merged in #91

PR #71, opened 2015-03-14
Removes shyp, adds node-pre-gyp.

tcr:


kevinmehall:
It looks like this publishes on every commit to master, but the filenames include only the version number, not the git hash. Will those overwrite the last release build?

tcr:
Yup. We could create artifacts, or publish on a commit tag that validates as a semver, or tags with "[release]" in them?

kevinmehall:
Tags would probably be best since those should happen anyway. Tags that match `^v[0-9]`?

PR #93, opened 2015-10-23
fix mac build

heavyk:
resolves my compiler error.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env sh
+
+node-pre-gyp install 2> /dev/null || \
+	([ -d "libusb" ] || git clone --depth=1 https://github.com/kevinmehall/libusb.git) && \
+	node-pre-gyp rebuild
--- a/binding.gyp
+++ b/binding.gyp
@@ -50,10 +50,10 @@
           }],
           ['OS=="mac"', {
             'xcode_settings': {
-              'OTHER_CFLAGS': [ '--std=c++1y' ],
+              'OTHER_CFLAGS': [ '--std=c++1y', '-stdlib=libc++' ],
               'OTHER_LDFLAGS': [ '-framework', 'CoreFoundation', '-framework', 'IOKit' ],
               'SDKROOT': 'macosx',
-              'MACOSX_DEPLOYMENT_TARGET': '10.5',
+              'MACOSX_DEPLOYMENT_TARGET': '10.7',
             },
           }],
           ['OS=="win"', {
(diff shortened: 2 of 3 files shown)


heavyk:
one sec. trying to figure out the submodule thing to install from git repo

heavyk:
ok, I got it. you can test it:
```
npm i 'heavyk/node-usb#fix-osx'
```

I remembered I need to check to see if the libusb dir doesn't exist before installing. cause I think libusb is included in the package.. I believe it's working now

install.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+
+node-pre-gyp install 2> /dev/null || \
+	([ -d "libusb" ] || git clone --depth=1 https://github.com/kevinmehall/libusb.git) && \
```
kevinmehall: `git submodule update --init --recursive`

heavyk: it doesn't work. try it. it just returns without doing anything. I did some research and everyone else is cloning

heavyk: I'll make another branch so you can see.. one sec

heavyk: ok, try it with: `npm i 'heavyk/node-usb#submodule-no-init'`

kevinmehall:
@tcr anything here we should be doing? Is there a better way to make npm get submodules when the git url is referenced by a package?

tcr:
@heavyk @kevinmehall

Sorry for the delay in responding. @heavyk, your change to add `-stdlib=libc++` has landed on master, I had to extract that for the 1.1.0 release. You can rebase to see that reflected in your code.

I believe this is the issue: the libusb source needs to be packaged with the published npm tarball. It's pretty easy for this to accidentally not happen, so I think the "prepublish" step needs a `git submodule update --init --recursive` step to ensure it.

On the other hand, the 1.1.0 release does have this. @heavyk, can you try running this on your OS X box:

```
npm install usb --build-from-source
```

If that works, that means we don't need the wrapper script around `./install.sh`, but just to ensure we're not publishing tarballs without libusb source.

heavyk:
@tcr - yeah that command will work, but

```
npm i nonolith/node-usb --build-from-source
```

does not -- which is what I was attempting to fix.

heavyk:
and from my tests, `git submodule update --init --recursive` will not work. it simply does nothing. you would need to clone it.

nevermind, don't worry about it

tcr:
@heavyk I pulled in your NPM update commit, since this is also needed. https://github.com/nonolith/node-usb/pull/99

Thanks for explaining your usecase, and indeed, it does not work. We should advocate for both https://github.com/npm/npm/issues/3055 or https://github.com/npm/npm/issues/1876 to be fixed. In the long term, it probably is best to keep these install changes isolated to a given development fork that's likely to be used by the Github install shorthand.

PR #100, opened 2015-10-31
Fix use-after-free in descriptor extra data (1.1.0 regression)

kevinmehall:
Nan::NewBuffer takes ownership of the passed buffer, but the
descriptors are freed once converted to JS. Use Nan::CopyBuffer
instead, and remove the wrapper function, as it no longer wraps
multiple V8 calls.

### DIFF
--- a/src/device.cc
+++ b/src/device.cc
@@ -11,10 +11,6 @@
 
 static Nan::Persistent<v8::FunctionTemplate> device_constructor;
 
-Local<Object> makeBuffer(const unsigned char* ptr, unsigned length) {
-	return Nan::NewBuffer((char*) ptr, (uint32_t) length).ToLocalChecked();
-}
-
 Device::Device(libusb_device* d): device(d), device_handle(0) {
 	libusb_ref_device(device);
 	DEBUG_LOG("Created device %p", this);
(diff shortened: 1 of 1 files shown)


tcr:
lgtm.

johnnyman727:
I don't believe this is the root cause of https://github.com/tessel/t2-firmware/issues/121.

My test procedure:
* inside the `node_modules` folder of `t2-cli`, `rm -rf usb` to get rid of existing library
* cloned this repo recursively and checked out this branch
* `npm install`
* renamed the directory from `node-usb` to `usb`
* ran `t2 update`
* still got the same error

CarlQLange:
Manually applying this patch and rebuilding from source fixed the issue for me.

johnnyman727:
@CarlQLange which issue did this cause exactly?

CarlQLange:
Couldn't update the firmware due to an invalid DFU descriptor, but I also forced the use of `usb 1.1.0` (not `1.0.6`) as the older version was otherwise broken on Node 5.

johnnyman727:
Edit: I had to rebuild this branch manually (`./node_modules/.bin/node-pre-gyp rebuild`) because it was pulling from remote. After rebuilding, it fixed the issue for Tessel 2 as well!

rwaldron:
@johnnyman727 re...

> After rebuilding, it fixed the issue for Tessel 2 as well!

Does that mean you're no longer hitting the invalid DFU descriptor issue?

rwaldron:
@johnnyman727 I saw your update on the other issue, so the answer is yes. This is great news!

johnnyman727:
:+1:

PR #113, opened 2016-02-16
libusb on linux kernel 4+ advice

reqshark:
@tcr, I'm having to do: 

```sh
$ sudo ln -s /sys/bus /dev/bus
``` 

to get libusb and node-usb to work w/ more recent linux kernels. This PR makes a note of the change.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -12,7 +12,7 @@ It's based entirely on libusb's asynchronous API for better efficiency, and prov
 Installation
 ============
 
-Libusb is included as a submodule. On Linux, you'll need libudev to build libusb. On Ubuntu/Debian: `sudo apt-get install build-essential libudev-dev`
+Libusb is included as a submodule. On Linux, you'll need libudev to build libusb. With Kernel v4+, symlink `/dev/bus` to its new location, `/sys/bus` to work with libusb. On Ubuntu/Debian: `sudo apt-get install build-essential libudev-dev`
 
 Then, just run
 


kevinmehall:
What distribution are you using? I have kernel 4.2.0 on Ubuntu 15.10 and don't have this issue.

The contents of /sys/bus and /dev/bus are not the same:
```
~ % ls /dev/bus/usb
001/  002/  003/
~ % ls /sys/bus/usb
devices/  drivers/  drivers_autoprobe  drivers_probe  uevent
```

reqshark:
mostly debian 8 but I also saw the need on my ubuntu.. plus amazon's normal ec2 ami for ubuntu 

```bash
18:24 $ lsb_release -a; uname -r; ls -la /dev/bus
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 15.10
Release:	15.10
Codename:	wily
4.2.0-040200-generic
total 0
drwxr-xr-x  3 root root   80 Feb 18 15:08 .
drwxr-xr-x 20 root root 4240 Feb 18 15:08 ..
lrwxrwxrwx  1 root root    8 Feb 18 15:08 bus -> /sys/bus
drwxr-xr-x  7 root root  140 Feb 18 15:08 usb
```
that's my output after i sym linked `/sys/bus`

checking `bus/usb` outputs (rather unexpectedly) same as above:

```bash
18:29 $ ls /dev/bus/usb
001  002  003  004  005
18:29 $ ls /sys/bus/usb
devices  drivers  drivers_autoprobe  drivers_probe  uevent
```

kevinmehall:
If `/dev/bus/` already exists, your `ln `command will create the link `/dev/bus/bus`, which you're seeing when you run `ls -la /dev/bus`. Libusb surely isn't looking there, so I don't think this had anything to do with the problem.

Note that most cloud VMs don't simulate USB controllers, so don't expect node-usb to work on ec2.

If you can reproduce this bug on a local machine, please open an issue with full debug log output (add `usb.setDebugLevel(4)` to your program).

reqshark:
ok thanks will do!

> Note that most cloud VMs don't simulate USB controllers, so don't expect node-usb to work on ec2.

that `ln -s` trick gets it working on the vm, which might be helpful for testing or for consistency

PR #126, opened 2016-06-12
Adding 6 to see if it compiles...

rwaldron:
This should be updated to use node_js key, but I really just want to see if this is building on these versions and I'm too lazy to do it manually


rwaldron:
Elsewhere... https://travis-ci.org/tessel/t2-cli/jobs/125957491

tcr:
Going to close this.

PR #132, opened 2016-08-02
Added parent property

lal12:
Added parent property to device object. If no parent exists property is null.


kevinmehall:
Merged as 60bc9eeab

PR #137, opened 2016-08-23
ci: add node.js 6

rwaldron:
Just opening this to get an idea of what the failures look like, this way we have a single source of truth to refer to

Signed-off-by: Rick Waldron <waldron.rick@gmail.com>


rwaldron:
That wasn't really my intention, but my intention was never for this PR to land either...


> Just opening this to get an idea of what the failures look like, this way we have a single source of truth to refer to

tcr:
Closing in favor of #138

PR #142, opened 2016-10-04
Simplifies the bindings include.

tcr:
This uses the node `bindings` module and changes the node-pre-gyp `module_path` to output to a directory the node `bindings` module can find.

This simplifies code that has to analyze where these bindings are found programmatically.


usb.js
```diff
@@ -1,8 +1,6 @@
-var binary = require('node-pre-gyp');
-var path = require('path');
-var binding_path = binary.find(path.resolve(path.join(__dirname,'./package.json')));
+var bindings = require('bindings')('usb_bindings.node');
 
-var usb = exports = module.exports = require(binding_path);
+var usb = exports = module.exports = bindings;
```
rwaldron: Since that `bindings` is never used after this, you can consolidate: 

```js
var usb = exports = module.exports = require('bindings')('usb_bindings.node');
```

tcr: Okay, didn't know if that screwed up the AST analysis heh

rwaldron:
LGTM

PR #154, opened 2016-11-26
Always use system libusb on FreeBSD

myfreeweb:
Upstream libusb does not support FreeBSD. (Even though it does support OpenBSD, NetBSD and even Haiku.)

All tests passed on my FreeBSD machine!

### DIFF
--- a/src/device.cc
+++ b/src/device.cc
@@ -81,14 +81,29 @@ static NAN_METHOD(deviceConstructor) {
 	STRUCT_TO_V8(v8dd, dd, bNumConfigurations)
 
 	uint8_t port_numbers[MAX_PORTS];
+	// FreeBSD workarounds based on https://sourceforge.net/p/sigrok/mailman/attachment/54C11DD2.10002%40uffe.org/1/
+#ifdef __FreeBSD__
+	struct libusb_device_handle* dh;
+	CHECK_USB(libusb_open(self->device, &dh));
+#endif
 	int ret = libusb_get_port_numbers(self->device, &port_numbers[0], MAX_PORTS);
+#ifdef __FreeBSD__
+	if (ret <= 0) {
+		Local<Array> array = Nan::New<Array>(1);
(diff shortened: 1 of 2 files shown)


myfreeweb:
Wait, it segfaults when listing devices as root (with access to `/dev/usb/*`)…

myfreeweb:
Okay, now everything works great :)

src/device.cc
```diff
@@ -81,14 +81,29 @@ static NAN_METHOD(deviceConstructor) {
...
+	// FreeBSD workarounds based on https://sourceforge.net/p/sigrok/mailman/attachment/54C11DD2.10002%40uffe.org/1/
+#ifdef __FreeBSD__
+	struct libusb_device_handle* dh;
+	CHECK_USB(libusb_open(self->device, &dh));
+#endif
 	int ret = libusb_get_port_numbers(self->device, &port_numbers[0], MAX_PORTS);
+#ifdef __FreeBSD__
+	if (ret <= 0) {
+		Local<Array> array = Nan::New<Array>(1);
+		array->Set(0, Nan::New(libusb_get_device_address(self->device)));
```
kevinmehall: `libusb_get_device_address` != `libusb_get_port_numbers`. I think it would be best fall back to not setting the property if not supported, rather than to fill it with another, unrelated number.

I assume the segfault is when it passes a negative `ret` to Nan::New. It should simply check that the result is >= 0 before using it, on all platforms.

myfreeweb: That's what was done [in the linked sigrok patch](https://sourceforge.net/p/sigrok/mailman/attachment/54C11DD2.10002%40uffe.org/1/).

kevinmehall:
Thanks! I've merged this without the `else` case.

If your program wants the device address if the port numbers are not available, then you can do that explicitly in JS: `dev.portNumbers || [dev.deviceAddress]`. You might want to include `dev.busNumber` in there too, or it won't even be guaranteed unique if there are multiple USB buses.

PR #209, opened 2017-12-02
Add Binary Object Store descriptor

thegecko:
This PR adds support for reading Binary Object Store descriptors for USB devices. The bosDescriptor can be used to detect extended capabilities of the device such as support for Wireless USB or platform-specific capabilities such as [WebUSB](https://wicg.github.io/webusb/#platform-descriptor).

It includes updates to the docs and tests, although without knowing the device connected for testing, I don't know whether tests will pass. Further tests can be added if the test device has valid BOS capabilities.

I haven't included capability decoders (these can be found in `Table 9-14` in the USB 3.2 spec.), but these could be added as a future PR if required.

The functionality uses `Nan::DefineOwnProperty` inline with recent changes, but let me know if anything else needs modifying.


tessel-highfive:
Thanks for the pull request, and welcome! The Rust team is excited to review your changes, and you should hear from @rwaldron (or someone else) soon.

kevinmehall:
Squashed as 4bdc7279a502caa6b264d503475d7005e2e5eb48. Thanks!

kevinmehall:
@thegecko I didn't review this carefully enough and I'm going to have to revert it. While `libusb_get_config_descriptor` simply returns cached data from the kernel, `libusb_get_bos_descriptor` is a blocking call that fetches the descriptor from the device, so it needs to be an async function, not a property getter. This library shouldn't cause bus traffic upon opening a device that was not explicitly requested.

Because it's not doing anything special in libusb, it's probably easier to implement this in JS on top of the existing async `controlTransfer` method like [getStringDescriptor](https://github.com/tessel/node-usb/blob/master/usb.js#L156) than to write C++ to run the libusb function in a thread pool.

Additionally, this request seems to be crashing [at least one device](https://github.com/tessel/node-usb/issues/211#issuecomment-350325695). It needs to check `this.deviceDescriptor.bcdUSB > 0x0201` before performing this transfer.

thegecko:
That's annoying. I'll check out the `controlTransfer` approach.

thegecko:
The async approach will require a function to be exposed which can't be easily massaged into the same form of other descriptor access methods.

I think the best approach will be to expose `getBosDescriptor()` function and also a `getCapabilities()` function which enumerates all capabilities from the BOS descriptor asynchronously on demand. I'll try to craft something which also incorporates the check as outlined by @rhuehn in #213.

PR #213, opened 2017-12-08
Check version before __getBosDescriptor (#211)

### LINKED ISSUE #211: V1.4 Interface.claim() fails with LIBUSB_ERROR_NO_DEVICE using USB 1.1 device
I upgraded from v1.2->v1.4 to fix build errors on node v9 and had the bad luck to find node-usb broke with my X10 CM15A device. A temporary workaround theatersoft/x10@59470a8 pins the working version before #209 PR.

Probable cause is a missed edge case with devices lacking a Binary Object Store descriptor, as my problem device is USB 1.1.

rhuehn:
This fixes #211 on my USB 1.1 device which disconnects on calls to libusb_get_device_descriptor.

Based on similar version checking in libusb examples e.g. https://github.com/libusb/libusb/blob/b4c9b4272d61cecffeddeb91abd31efe256a6224/examples/testlibusb.c#L236

### DIFF
--- a/usb.js
+++ b/usb.js
@@ -62,6 +62,9 @@ Object.defineProperty(usb.Device.prototype, "configDescriptor", {
 
 Object.defineProperty(usb.Device.prototype, "bosDescriptor", {
 	get: function() {
+		if (this.deviceDescriptor.bcdUSB < 0x201){
+			return null;
+		}
 		try {
 			return this._bosDescriptor || (this._bosDescriptor = this.__getBosDescriptor())
 		} catch(e) {


tessel-highfive:
Thanks for the pull request, and welcome! The Rust team is excited to review your changes, and you should hear from @rwaldron (or someone else) soon.

kevinmehall:
Thanks for this. However, I think the BOS support is going to need to be largely rewritten because `libusb_get_bos_descriptor` is documented to be a blocking API, thus it should be an async function rather than a property getter. I agree that any replacement implementation should include this check.

For now, I've deprecated 1.4.0 and tagged 1.3.1 without the BOS commit, and will publish it as `latest` once the Win/Mac binaries build on CI.

Issue #147, opened 2016-10-29
LIBUSB_ERROR_NOT_SUPPORTED problem on Windows


fjozsef:
Hi!

I have the following algorithm:
```node
#!/usr/bin/env node
'use strict';

var usb = require('usb');

// replaced with the correct ids
var vid = 0x0000;
var pid = 0x0000;

var device = usb.findByIds(vid, pid);
device.open();

var usbInterface = device.interface(0);
if (usbInterface.isKernelDriverActive()) {
    usbInterface.detachKernelDriver();
}
```

The problem is that at `usbInterface.isKernelDriverActive()` line throws on windows 10, but it works perfectly on Linux ubuntu.
Error:

```bash
C:\work\usb\node_modules\usb\usb.js:206
        return this.device.__isKernelDriverActive(this.id)
                           ^

Error: LIBUSB_ERROR_NOT_SUPPORTED
    at Error (native)
    at Interface.isKernelDriverActive (C:\work\usb\node_modules\usb\usb.js:206:21)
```

I've downloaded Zadig 2.2 from [here](http://zadig.akeo.ie/) and installed WinUSB, but the problem persisted.

Could you help what I am doing wrong?

kevinmehall:
http://libusb.sourceforge.net/api-1.0/group__dev.html#gab14d11ed6eac7519bb94795659d2c971
> This functionality is not available on Windows.

You probably don't need the isKernelDriverActive / detachKernelDriver even on Linux, as most devices you'd use libusb for don't have kernel drivers.

sairam1919:
Hi ,

I have the following code

var usb = require('usb');
term = usb.findByIds(1921,21904);
console.log(term);
term.open();

when i try to run the code i am getting the following error

  this.__open()
             ^

Error: LIBUSB_ERROR_NOT_SUPPORTED
    at Error (native)
    at Device.usb.Device.open (C:\Users\PO323106\Desktop\unity_poc\node_modules\usb\usb.js:33:7)
    at Object.<anonymous> (C:\Users\PO323106\Desktop\unity_poc\usb.js:10:6)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:394:7)

what is the issue with this.

[...]

PR #227, opened 2018-01-24
Do not call kernel driver functionality when running on win32

IvanSanchez:
I've been hit by #147, and I think that a nicer approach would be to just skip calls to `__isKernelDriverActive`, `__detachKernelDriver` and `__attachKernelDriver` when running on a win32 platform.

This PR does that, and makes a brief note in the readme.

(The use case is a [JS implementation of USB CDC ACM](https://www.npmjs.com/package/usb-cdc-acm) which, in fact, has to detach the `cdc-acm` kernel driver)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -159,13 +159,13 @@ Releases the interface and resets the alternate setting. Calls callback when com
 It is an error to release an interface with pending transfers. If the optional closeEndpoints parameter is true, any active endpoint streams are stopped (see `Endpoint.stopStream`), and the interface is released after the stream transfers are cancelled. Transfers submitted individually with `Endpoint.transfer` are not affected by this parameter.
 
 ### .isKernelDriverActive()
-Returns `false` if a kernel driver is not active; `true` if active.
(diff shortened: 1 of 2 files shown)


tessel-highfive:
Thanks for the pull request, and welcome! The Rust team is excited to review your changes, and you should hear from @hipsterbrown (or someone else) soon.

kevinmehall:
@ivankennethwang: that's not the same error. LIBUSB_ERROR_NOT_SUPPORTED means an incompatible driver is currently bound to the device. Replace the driver with WinUSB using http://zadig.akeo.ie/.

ivankennethwang:
@kevinmehall  I did try this one out however it failed to upgrade my drivers. Also, I notice that the problem only happens with node-usb being rebuilt on an electron project. If I create a regular project with node-usb, my application can actually connect to the USB device properly. That's why I'm a bit confused whether it is really the drivers or not that is causing my issue.

kevinmehall:
Any code that needs this is going to be platform-specific anyway, so making it a no-op I think just makes it more confusing.

PR #241, opened 2018-04-18
add endpoint.clearHalt

guenterh22:
Hi,

I need the clearHalt request from libusb in my nodejs project. I have added the required code. It is tested with some USB mass storage devices and validated with USB protocol analyzer on EHC host controller. It would be great if you can accept the request. For question please contact me at guenter.hildebrandt@emsys.de. 

Regards
Guenter Hildebrandt

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -208,6 +208,10 @@ Object with fields from the endpoint descriptor -- see libusb documentation or U
 ### .timeout
 Sets the timeout in milliseconds for transfers on this endpoint. The default, `0`, is infinite timeout.
 
+### .clearHalt()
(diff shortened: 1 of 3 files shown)


hauptmedia:
I just stumpled upon this pull request. I didn't see it and implemented the same thing in PR #275.
The libusb documentation states that libusb_clear_halt is a blocking function so I think handling it in the libuv queue (like libusb_reset_device) seems to be a better solution because otherwise this could lead to a block in nodejs.

kevinmehall:
Sorry I never reviewed this, but @hauptmedia is correct. I merged #275.

PR #243, opened 2018-05-03
Call ref() after CHECK_USB in Transfer_Submit

### LINKED ISSUE #242: Missing unrefs in Transfer_Submit if libusb_submit_transfer fails.
Here https://github.com/tessel/node-usb/blob/master/src/transfer.cc#L87

If `libusb_submit_transfer` fails, the `CHECK_USB` macro throws an error and `self->unref()` and `self->device->unref()` are not called preventing the device from being closed later with a `Can't close device with a pending request` error.

zvin:
This fixes #242


zvin:
Sorry about the branch name, it should have been `move-refs-after-check-usb`.

kevinmehall:
Good find, thanks!

PR #253, opened 2018-07-26
Remove Node versions 5 and 7 from CI

jrobeson:
Both are EOL and in distro's supported releases.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,9 +9,7 @@ environment:
 
   matrix:
     - node_version: '4'
-    - node_version: '5'
     - node_version: '6'
-    - node_version: '7'
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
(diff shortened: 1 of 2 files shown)


jrobeson:
It looks like 9 just passed EOL as 6-30-2018, so that could go too if you wish.

appveyor.yml
```diff
@@ -9,9 +9,7 @@ environment:
 
   matrix:
     - node_version: '4'
-    - node_version: '5'
     - node_version: '6'
-    - node_version: '7'
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
       GYP_MSVS_VERSION: 2015
```
kevinmehall: These lines also need to be removed or they'll apply to node 6 above

kevinmehall:
Went ahead and removed these (as well as 4 and 9) in 4a96a14

PR #260, opened 2018-08-17
Allow loading module in environments without usb

### LINKED ISSUE #194: "Could not initialize libusb. Check that your system has a usb controller." on Travis CI
For some reason I get that error only when running inside a Travis CI job.

I tracked down the error to `libusb_init` (http://libusb.sourceforge.net/api-1.0/group__lib.html#ga9517c37281bba0b51cc62eba728be48b). Do you have any clues about why that's failing, and how we can at least workaround it?

webmaster128:
e.g. CI system.

This throws an exception when USB is not available only when used, not when module is loaded.

Check currently happens

* in findByIds()
* in Device.open()
* in Interface constructor
* in Endpoint constructor

Please let me know if those places should be reduced or expanded. I am not an expert of USB.

Closes #194

### DIFF
--- a/src/node_usb.cc
+++ b/src/node_usb.cc
@@ -84,8 +84,8 @@ extern "C" void Initialize(Local<Object> target) {
 	Device::Init(target);
 	Transfer::Init(target);
 
-	Nan::SetMethod(target, "setDebugLevel", SetDebugLevel);
-	Nan::SetMethod(target, "getDeviceList", GetDeviceList);
+	Nan::SetMethod(target, "_setDebugLevel", SetDebugLevel);
+	Nan::SetMethod(target, "_getDeviceList", GetDeviceList);
 	Nan::SetMethod(target, "_enableHotplugEvents", EnableHotplugEvents);
 	Nan::SetMethod(target, "_disableHotplugEvents", DisableHotplugEvents);
 	initConstants(target);
--- a/usb.js
+++ b/usb.js
@@ -6,9 +6,11 @@ var usb = exports = module.exports = require(binding_path);
 var events = require('events')
 var util = require('util')
 
-// Check that libusb was initialized.
-if (usb.INIT_ERROR) {
-	throw new Error('Could not initialize libusb. Check that your system has a usb controller.');
+function ensureLibusbInitialized() {
+	// Check that libusb was initialized.
+	if (usb.INIT_ERROR) {
+		throw new Error('Could not initialize libusb. Check that your system has a usb controller.');
+	}
 }
 
 Object.keys(events.EventEmitter.prototype).forEach(function (key) {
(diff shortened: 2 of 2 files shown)


HipsterBrown:
Thank for creating this PR! 🎉 

I like this update because I have been annoyed by `node-usb` throwing when I'm not using any of the methods in `t2-cli`. Looking for feedback from @rwaldron and @tcr as well.

kevinmehall:
These checks should probably happen in C++ instead of in JS. For one, `usb.getDeviceList` needs to be protected and is directly exposed from C++.

The Endpoint and Interface constructors, and Device.open don't really need it, since they can't do anything without a `Device`, and nothing can construct a `Device` without a USB controller.

`usb._enableHotplugEvents` and `usb._disableHotplugEvents` should have a check in C++ as well. It might be surprising for `usb.on('attach') to throw, so maybe we should wrap both of the `exports.on` calls at the bottom of usb.js in a `if (!usb.INIT_ERROR) {` such that you can still listen for hotplug events without a USB controller, but you'll just never get any.

webmaster128:
Isn't that whole check a convenience feature to create a better error message instead of something cryptic when USB functionality fails? In this case I should not be an issue if a user is able to work around that check as long as it is performed at least once in all mayor use cases.

I don't have experience in throwing exceptions from C++ that can be caught in JavaScript, so if we need C++ checks, I'd like to ask someone else to take over.

Regarding getDeviceList: good point. Can we wrap that in JS in a backwards compatible way? E.g. by exposing _getDeviceList from C++.

> since they can't do anything without a Device, and nothing can construct a Device without a USB controller.

okay and `Device`s are constructed in getDeviceList and findByIds?

kevinmehall:
No, it will use an uninitialized libusb_context and corrupt memory, or if you're lucky, segfault.

> okay and Devices are constructed in getDeviceList and findByIds?

getDeviceList and the hotplug events. `findByIds` just calls getDeviceList.

kevinmehall:
Actually the C++ module already bails out on initialization failure before even exposing any of the problematic methods to JS. So the JS code just has to be able to deal with everything on `usb` being undefined, and I don't think you have to touch the C++.

kevinmehall:
Looks good, thanks!

kevinmehall:
It doesn't actually work since `usb.Device` is undefined.

```
% docker run -v $PWD:/pwd -it node:10
> require('/pwd/usb')
TypeError: Cannot read property 'prototype' of undefined
    at Object.<anonymous> (/pwd/usb.js:42:12)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
```

I pushed a commit with an alternate approach: https://github.com/tessel/node-usb/commit/d0d7abaeb7ac0c50bfa7af2e70699c4934d18dc8

PR #276, opened 2018-10-27
fix endpoint polling status check

### LINKED ISSUE #249: Polling Start/Stop on Input Endpoint
Hello,

I have some trouble understanding how Polling Start / Stop mechanism works.

```js
const handler = dat => {
    console.log(dat);
    epIn.stopPoll(() => {
        epIn.startPoll(1);
    });
};
epIn.on('data', handler);
epIn.startPoll(1);
```

  * attaching event handler `.on('data', ...)`
  * starting polling `.startPoll(1)`
  * getting callback with the data
  * stopping polling `.stopPoll(cb)` with callback to re-enable polling.
  * and getting Error:

```js
/home/drom/work/lab/drom/ecp5/node_modules/usb/usb.js:268
                throw new Error("Polling already active")
                ^

Error: Polling already active
    at InEndpoint.Endpoint.startPoll (/home/drom/work/lab/drom/ecp5/node_modules/usb/usb.js:268:9)
    at InEndpoint.startPoll (/home/drom/work/lab/drom/ecp5/node_modules/usb/usb.js:324:61)
    at InEndpoint.epIn.stopPoll (/home/drom/work/lab/drom/ecp5/lib/ftdi.js:99:22)
    at Object.onceWrapper (events.js:273:13)
    at InEndpoint.emit (events.js:182:13)
    at Transfer.transferDone (/home/drom/work/lab/drom/ecp5/node_modules/usb/usb.js:340:10)
```
How come it is still active?
Do I need to clean data buffer after `.on('data', ...)` callback?

hauptmedia:
If you call startPoll, stopPoll & startPool on an endpoint you will always get a "Polling already active" exception because the polling state is not properly checked.

This resolves #249

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -238,7 +238,7 @@ is called once all transfers have completed or canceled.
 Emitted with data received by the polling transfers
 
 ### Event: error(error)
-Emitted when polling encounters an error.
+Emitted when polling encounters an error. All in flight transfers will be automatically canceled and no further polling will be done. You have to wait for the `end` event before you can start polling again.
 
 ### Event: end
 Emitted when polling has been canceled
--- a/usb.js
+++ b/usb.js
@@ -342,6 +342,7 @@ InEndpoint.prototype.startPoll = function(nTransfers, transferSize){
 			self.pollPending--
 
 			if (self.pollPending == 0){
+				delete self.pollTransfers;
 				self.emit('end')
 			}
 		}


kevinmehall:
If you `stopPoll` and then immediately `startPoll`, the async cancellation of the first set of transfers will still be pending. In their completion callback they will see `pollActive` as `true` and be resubmitted. But those transfers aren't in `this.pollTransfers`, so another `stopPoll` won't cancel them.

It might be better to clear `pollTransfers` before emitting `'end'` instead of changing these checks to `pollActive`, so that you can't restart polling until all transfers have been cancelled.

hauptmedia:
Ok, looking at the code again your described functionality should already work if you do not call startPoll again until the callback from stopPoll fired. But my application will still crash after a while, will do further investigations here.

hauptmedia:
I found my error: i immediately restarted polling after an error occured. But there can be still transfers in flight that will be automatically canceled due to the error.

I changed this PR to add a few words about this behavior in the docs.

hauptmedia:
I also added a commit for clearing `pollTransfers` if all in-flight transfers are canceled.

kevinmehall:
Thanks!

PR #311, opened 2019-05-07
Making module work with Node 12

imrehg:
The changes required 2 parts:

* the first commit needed updates to the abstraction layer, as there are a number calls that are now deprecated on Node 12. Updated the code to use Nan properly for those lines as well, and thus work for all Node versions.
* the very old mocha didn't work on Node 12 (did not run any tests), thus bumping version and update the test calls allows tests work on all Node versions as well.

As an extra, also updated Travis config building for Node 12.0.0 as well.

### DIFF
--- a/src/node_usb.cc
+++ b/src/node_usb.cc
@@ -96,11 +96,11 @@ NODE_MODULE(usb_bindings, Initialize)
 
 NAN_METHOD(SetDebugLevel) {
 	Nan::HandleScope scope;
-	if (info.Length() != 1 || !info[0]->IsUint32() || info[0]->Uint32Value() > 4) {
+	if (info.Length() != 1 || !info[0]->IsUint32() || Nan::To<v8::Uint32>(info[0]).ToLocalChecked()->Value() > 4) {
 		THROW_BAD_ARGS("Usb::SetDebugLevel argument is invalid. [uint:[0-4]]!")
 	}
 
-	libusb_set_debug(usb_context, info[0]->Uint32Value());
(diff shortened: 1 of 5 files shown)


src/node_usb.cc
```diff
@@ -297,7 +297,6 @@ void initConstants(Local<Object> target){
 
 Local<Value> libusbException(int errorno) {
 	const char* err = libusb_error_name(errorno);
-	Local<Value> e  = Nan::Error(err);
-	e->ToObject()->Set(Nan::New<String>("errno").ToLocalChecked(), Nan::New<Integer>(errorno));
+	Local<Value> e  = Nan::ErrnoException(errorno, NULL, err, NULL);
```
kevinmehall: Does `ErrnoException` assume that the value passed is a standard system error code? These are [`libusb_error`](http://libusb.sourceforge.net/api-1.0/group__libusb__misc.html#gab2323aa0f04bc22038e7e1740b2f29ef) or [`libusb_status`](http://libusb.sourceforge.net/api-1.0/group__libusb__asyncio.html#ga9fcb2aa23d342060ebda1d0cf7478856) values.

imrehg: @kevinmehall very good point, I wasn't sure about this part. I've updated the code to keep closer to the original one, and as much as I can tell, it retains the original behaviour, while works on Node 12. I don't have the demo device that is being tested with `full-test` in this repo, but as much as I poked around, it seems to match up.

Does it look better?

kevinmehall:
Passed `full-test` on Node v12.2.0 on Linux.

Thanks!

PR #334, opened 2019-10-12
Adds specified Buffer implementation

### LINKED ISSUE #333: Control Transfer Buffer checking is faulty
The current implementation of `controlTransfer` relies on passing in a `Buffer` for and then uses the `Buffer.isBuffer` method for validation.

https://github.com/tessel/node-usb/blob/0d3c0e4694f52760c9b3f7ade0abb0f40ad50d03/usb.js#L104-L106

The problem is that the NodeJS `Buffer` shim calls a function which does an `instanceOf` comparison, which requires that the `Buffer` being passed in is utilizing the same library or else the validation fails, even if the implementations are compatible.  This leads to some very strange side-effects when working with systems that have different bundling processes.  In webpack, `node` targeted bundles will pull in a different shim than `web` targeted bundles which will cause issues if you are trying to make calls between the two build targets (as is often the case with Electron applications).

A possible resolution is to include a `Buffer` implementation library.  This has the advantage of solving this initial problem because the `isBuffer` method in the [feross/buffer](https://github.com/feross/buffer) library is more robust and won't fail if the `Buffer` object is not using the same implementation library.  The other advantage is that this allows users of the library to pass in a `Buffer` of the exact same library implementation regardless of build platform and reduces the possibility of API mismatches.

The downside to this approach is an added dependency, which for a project with only 3 dependencies is a non-trivial increase.  The other potential solution would be to implement a more robust validation method similar to [feross/is-buffer](https://github.com/feross/is-buffer).

I will submit a PR with the added `Buffer` library attached to this issue for testing and review.

bayssmekanique:
Resolves #333

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,27 @@
+{
+  "name": "usb",
+  "version": "1.6.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "base64-js": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.3.1.tgz",
+      "integrity": "sha512-mLQ4i2QO1ytvGWFWmcngKO//JXAQueZvwEKtjgQFM4jIK0kU+ytMfplL8j+n5mspOfjHwoAg+9yhb7BwAHm36g=="
+    },
+    "buffer": {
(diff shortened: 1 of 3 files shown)


kevinmehall:
There's no reason to use a Buffer polyfill in Node. Added a comment on #333 suggesting a better solution to the problem.

PR #339, opened 2019-11-07
[work in progress] Add arm64 prebuilds to Travis build

JamesHagerman:
Compiling node-usb on Raspberry Pi is not all that friendly. Prebuilds would be a useful addition if Travis plays along.

This is a work in progress. I'm still sorting out how to test Travis builds from a fork.


JamesHagerman:
Closing this until I can test it more. Travis is not very intelligent about excluding things like `osx` builds for `arm64`. :roll_eyes:

Issue #343, opened 2019-11-17
Cannot use node-usb with electron


GoogleFeud:
I get this error when I try running my main script with electron: `electron Engine/Engine.js`:

```
App threw an error during load
Error: A dynamic link library (DLL) initialization routine failed.
\\?\D:\Programs\Project\node_modules\usb\build\Release\usb_bindings.node
    at Error (native)
    at process.module.(anonymous function) [as dlopen] (ELECTRON_ASAR.js:173:20)
    at Object.Module._extensions..node (module.js:583:18)
    at Object.module.(anonymous function) [as .node] (ELECTRON_ASAR.js:173:20)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
    at bindings (D:\Programs\Pi-RetroConsole.js\node_modules\bindings\bindings.js:112:48)
```
Tried rebuilding the module, it said that the rebuild was successful, but the error still occurs. 

**Windows**: 10

**Dependencies**:
```
  "dependencies": {
    "electron": "^7.1.1",
    "usb": "^1.6.1",
    "ws": "^7.2.0"
  },
  "devDependencies": {
    "electron-rebuild": "^1.8.6"
  }
```

bayssmekanique:
I was receiving a similar error and it appears to be an issue with `electron-rebuild`.  For us, the solution was to switch to using `electron-builder` using the script `electron-builder install-app-deps` instead of the rebuild script.

quinton-ashley:
prebuild-install http 404 https://github.com/tessel/node-usb/releases/download/v1.6.3/usb-v1.6.3-electron-v82-win32-x64.tar.gz
    prebuild-install WARN install No prebuilt binaries found (target=10.1.3 runtime=electron arch=x64 libc= platform=win32)

thegecko:
Closing due to inactivity, please reopen this issue if it persists with the latest release.

PR #344, opened 2019-11-27
Update prebuild version

ivannkf:
Explanation from the prebuild repo:

Since electron and node are reusing the versions now (fx 6.0.0) and node-gyp only uses the version to store the dev files, they have started clashing. This new version of prebuild has a fix for it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   "devDependencies": {
     "coffeescript": "~2.4.1",
     "mocha": "~6.1.4",
-    "prebuild": "^8.2.1"
+    "prebuild": "^9.1.1"
   },
   "license": "MIT"
 }


kevinmehall:
Thanks! Does this explain #343?

ivannkf:
I'm not sure. The error I see is different, it just says that the node module version is the wrong one. I need version 73 (Electron 6) and I get version 48 (Node 6)

quinton-ashley:
prebuild-install http 404 https://github.com/tessel/node-usb/releases/download/v1.6.3/usb-v1.6.3-electron-v82-win32-x64.tar.gz
    prebuild-install WARN install No prebuilt binaries found (target=10.1.3 runtime=electron arch=x64 libc= platform=win32)

PR #345, opened 2019-11-28
Update libusb to latest version

ivannkf:
I saw you have a fork with hotplug support for windows. I don't think that's added in the original version. I needed some of the changes of the new library though, so here's the PR in case you want to merge it.


brandonros:
i would highly recommend merging this because windows 10 is broken for 32-bit, 64-bit node.js v10 + v12

just straight heap failures/seg faults, etc.

thegecko:
I'm reluctant to switch to the main libusb until hotplug support has been added. Far from ideal as we are missing a ton of patches I know.

PR #374, opened 2020-07-25
Adds NodeJS 14 to prebuild target and removes 13

MicahZoltu:
NodeJS 13 is now official end of life and NodeJS 14 is officially Current.  Looking at the prebuild target history, it appears that odd numbered versions (beta builds) aren't maintained so I dropped NodeJS 13 support and added 14 support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ addons:
 env:
   global:
     # Supported Node versions: https://nodejs.org/en/about/releases/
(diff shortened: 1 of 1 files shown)


MicahZoltu:
What is the release process like for this project?  I currently have a library that I cannot use since updating to NodeJS 14 due to the lack of prebuilt binaries and that my machine doesn't have the tooling necessary to build from scratch, and I would like to be able to use it again.  :)

PR #375, opened 2020-08-01
Bumps packages version in preparation for release.

MicahZoltu:
I'm hoping that CI is setup for this project and if this PR is merged then it will auto-release.  I'm currently blocked because I don't have the tooling required to compile this package from source on a target machine that is running NodeJS 14.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -2,6 +2,8 @@
   'variables': {
     'use_udev%': 1,
     'use_system_libusb%': 'false',
+    'v8_enable_pointer_compression': 0,
+    'v8_enable_31bit_smis_on_64bit_arch': 0,
   },
   'targets': [
     {
--- a/libusb.gypi
+++ b/libusb.gypi
@@ -1,6 +1,8 @@
 {
   'variables': {
     'use_udev%': 1,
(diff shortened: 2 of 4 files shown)


MicahZoltu:
Any progress on this?  Is there anything else I can do to assist with getting a new version out with NodeJS 14 support?

MicahZoltu:
I have made the necessary changes to get NodeJS 14 prebuilds working in Travis CI.  My hope is that this was what was keeping this PR from being merged and a new version deployed and now with these changes we can get a new version deployed.

Alternatively, if the build artifacts are merely added to GitHub then a new deploy is not needed.

MicahZoltu:
Note to future readers: I have a fork of this project at `@zoltu/usb` which implements this change and thus works in NodeJS 14.  I have no real intention of maintaining it, but if all you need is to just get things working in NodeJS 14 then you can swap out with that version as a temporary work around until hopefully this gets merged and deployed at some point.

thegecko:
Thanks @MicahZoltu 

The latest release moves to using Node-API, so it should now work with all Node.js versions >8

PR #377, opened 2020-08-26
allow accessing interfaces before setting a config

mildsunrise:
Hi! I'm porting a library that needs to access interfaces (to detach kernel drivers, etc) before calling `device.setConfiguration()`.
`usb` currently doesn't make that possible... I'd suggest a fix like this one.

### DIFF
--- a/usb.js
+++ b/usb.js
@@ -36,7 +36,6 @@ usb.Device.prototype.timeout = 1000
 
 usb.Device.prototype.open = function(defaultConfig){
 	this.__open()
-	if (defaultConfig === false) return
 	this.interfaces = []
 	var len = this.configDescriptor ? this.configDescriptor.interfaces.length : 0
 	for (var i=0; i<len; i++){


mildsunrise:
This is also needed to claim interfaces without setting a configuration.  
As indicated by [libusb documentation](http://libusb.sourceforge.net/api-1.0/libusb_caveats.html#configsel), setting configuration causes a lightweight device reset, which may be undesirable.

thegecko:
Thanks @mildsunrise. Would you mind outlining a test scenario I can use to try this with?

usb.js
```diff
@@ -36,7 +36,6 @@ usb.Device.prototype.timeout = 1000
 
 usb.Device.prototype.open = function(defaultConfig){
 	this.__open()
-	if (defaultConfig === false) return
```
thegecko, line 39: I think the logic here may be backwards and the solution is to change this to:

```javascript
if (defaultConfig === true) return
```

thegecko (changes requested):

thegecko:
Bump!

thegecko:
@mildsunrise is this still something you want attention on?

mildsunrise:
Hey sorry about the silence, I just don't know what to do with this.
This PR attempts to fix some of the issues detailed in #381, but ultimately the problem lies in the API.
Things have changed since I opened this, we now offer the WebUSB API which does not have these issues.

So I think a better approach is to make WebUSB work correctly (right now it doesn't seem to, since it uses the legacy API underneath) and then deprecate (or even freeze) the legacy API. @kevinmehall's comments in #401 seem to point in that direction.

thegecko:
> and then deprecate (or even freeze) the legacy API

I was wondering if the legacy API should literally be the exposure of choice libusb functions?

One drawback to removing this API is the loss of the non-blocking usb access this allows.

mildsunrise:
> exposure of choice libusb functions

what does 'choice libusb functions' mean? you mean the bits that are not exposed by WebUSB?

> One drawback to removing this API is the loss of the non-blocking usb access this allows

I'm also having trouble following here, isn't WebUSB also a non-blocking API?

thegecko:
> what does 'choice libusb functions' mean? you mean the bits that are not exposed by WebUSB?

I mean a raw interface that resembles common, low-level libusb calls.

> I'm also having trouble following here, isn't WebUSB also a non-blocking API?

I believe many of the legacy usb calls (e.g. `getDeviceList()`) return synchronous data from the kernel cache. Calls such as requests for strings need to interrogate the device so are async/blocking.

As WebUSB needs a lot of data loaded up to satisfy the device model, there's a lot more device access.

mildsunrise:
Ah, I see. Well... if we could go back in the past, I do agree that the legacy API should've just been a raw interface to libusb, with very little "magic".

But well, we ended up with this, and I think adding it now (either as a 3rd API, or by removing the current legacy API) would confuse users more. Maybe we could extend WebUSB with the libusb-specific bits it lacks? WebUSB looks close to a raw interface, we'd only have to give the option to disable the 'high-level' behaviour (basically the loading of string descriptors?)

thegecko:
OK, lets keep the legacy API (perhaps it should be called `raw` API?). I'm reluctant to extend the WebUSB API too much as it's already a standard and introduces incompatibility with the browser implementations.

> ...a better approach is to make WebUSB work correctly (right now it doesn't seem to...

Would you mind outlining how this doesn't work in a little more detail. Perhaps expanding #381 a little more with specific cases?

thegecko:
I'll close this for now and attack #381 issues considering the WebUSB API too.

PR #385, opened 2020-09-28
Add ref/unref methods for hotplug events

gfcittolin:
Hi there

This PR adds a simple functionality of calling `ref()`/`unref()` of the hotplug event queue from the user space. This allows the user code to listen to devices being attached and detached, but still allowing the process to gracefully exit without the need to manually remove the event listeners. 

This is very useful when user code is part of other bigger libraries and is not able to control when the process needs to terminate, so that the event listeners could be removed.

The following sample code can demonstrate the functionality:

```js
const usb = require('.');

usb.on('attach', dev => console.log(` >> ATTACH ${dev.busNumber}/${dev.deviceAddress}`));
usb.on('detach', dev => console.log(` >> DETACH ${dev.busNumber}/${dev.deviceAddress}`));
console.log('listening'); 

let tmr = setTimeout(() => {
    console.log('timer fired');
    usb.unrefHotplugEvents();
    //process should exit gracefully here
}, 5000);
tmr.unref(); //so that the only thing keeping the event loop alive is hotplug
console.log('timer set and unrefed');
```


thegecko:
Thanks @gfcittolin , would you mind updating your PR to use Node-API now we have made this switch?

thegecko:
Since `ref`/`unref` were removed from the queue when it was [ported to napi](https://github.com/node-usb/node-usb/commit/4cd686550de9732c78fb0fa5ba9d18b6e0233ad3), I think this PR is now redundant.

Please reopen if you disagree or this needs doing in another way.

thegecko:
I think a solution to your issue may be even simpler.

As the events extend the [Node.js event emitter](https://nodejs.org/dist/latest-v16.x/docs/api/events.html#class-eventemitter), you can use `removeAllListeners(...)` to achieve the same outcome, e.g.:

```javascript
const usb = require('.');

usb.on('attach', dev => console.log(` >> ATTACH ${dev.busNumber}/${dev.deviceAddress}`));
usb.on('detach', dev => console.log(` >> DETACH ${dev.busNumber}/${dev.deviceAddress}`));
console.log('listening'); 

let tmr = setTimeout(() => {
    console.log('timer fired');
    usb.removeAllListeners();
    //process should exit gracefully here
}, 5000);
tmr.unref(); //so that the only thing keeping the event loop alive is hotplug
console.log('timer set and unrefed');
```

gfcittolin:
Ok, I have to admit the original example was not the best one to demostrate the need for the use case. For that one, your solution makes much more sense.

But maybe think about this one:
```js
const usb = require('.');

// ... deep inside some library ...
function setupUsb() {
  usb.on('attach', dev => console.log(` >> ATTACH ${dev.busNumber}/${dev.deviceAddress}`));
  usb.on('detach', dev => console.log(` >> DETACH ${dev.busNumber}/${dev.deviceAddress}`));
  usb.unrefHotplugEvents();
  console.log('listening to hotplug events'); 
}

// ... in the user code ...
console.log('initializing user code');
setupUsb();
setTimeout(() => {
    console.log('timer fired, job done');
    // no ref left on user's control, process should exit gracefully here
}, 5000);
console.log('user code initialized');
```

Without `unrefHotplugEvents()`, the event loop is kept "live" and does not exit after the timer being fired.

The concept is to have the possibility to have an event to be triggered, but weakly referenced, so that it does not prevent the event loop to exit. Same idea of `unref()`ing a timer created with `setTimeout()`.

More specifically, I'll be using this inside a library on a [Node-RED](https://nodered.org/) node. The library dynamically creates class instances based on `attach` events that won't be tied to the flow execution (and therefore not tied to the flow start/stop events), but still needs to allow Node-RED to cleanly exit regardless of us listening to `attach` events. Let me know if you can think about another solution for this use case. 

Thanks for analyzing it!

PR #396, opened 2020-12-07
Added node-gyp options for node14 ci-build 

jins-tkomoda:
I found that there is no Node.js 14 prebuild binary(travis says 'passed' but actually 'failed' only node.js 14). This build problem occurs when building Node14 with older Nodejs versions. I added two options ('v8_enable_pointer_compression': 0, 'v8_enable_31bit_smis_on_64bit_arch': 0,) to binding.gyp and libusb.gypi.

In addition for future V8 compatibility, the [PR#394](https://github.com/tessel/node-usb/pull/394) is needed to apply.
But current travis-ci contaner use ubuntu trusty, that has old g++ that doesn't support c++14. So I upgrade trusty(14.04) to bionic(18.04). The reason why I didn't upgrade to focal(20.04) is that, focal is based on the Python3 and therefore there is more to fix. I will make another PR for that issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 - osx
 - windows
 
-dist: trusty
+dist: bionic
 
 addons:
   apt:
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,5 +1,7 @@
 {
   'variables': {
+		'v8_enable_pointer_compression': 0,
+		'v8_enable_31bit_smis_on_64bit_arch': 0,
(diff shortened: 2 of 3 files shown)


thegecko:
@jins-tkomoda we have now switched to using [a GitHub action](https://github.com/tessel/node-usb/blob/master/.github/workflows/prebuild.yml) to create the prebuild binaries instead of travis.

This includes a prebuild for node 14 which should appear on the next release.

PR #397, opened 2020-12-07
Python3 compliant for ci build

jins-tkomoda:
This PR migtates travis ci build enviroment from deprecated python2 to python3. Followings are changed.

- To use pyhon3 for default, simplly I changed ubuntu version to focal(20.04). I think it will work on older ubuntu distributions, but changing the default python to python3 makes the build script more complex.
- Upgrade npm that requires older node-gyp which doesn't support python3. This operation requires root privilage.
- Upgrade rebuild that requires older node-gyp which doesn't support python3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 - osx
 - windows
 
-dist: trusty
+dist: bionic
 
 addons:
   apt:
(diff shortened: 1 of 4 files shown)


thegecko:
@jins-tkomoda, we have now switched to using [a GitHub action](https://github.com/tessel/node-usb/blob/master/.github/workflows/prebuild.yml) to create the prebuild binaries instead of travis.

Is it OK to close this PR?

PR #403, opened 2021-01-13
Migrate to GitHub actions from travis

thegecko:
With reference to https://github.com/tessel/node-usb/issues/401, this PR migrates the prebuild CI functionality from travis to GitHub actions.

Changes of note:

- Existing OS/Arch matrix has persisted
- Node 6.x builds have been dropped to allow use of the MacOS VM on GH (10.13 isn't supported)
- `prebuild` has been updated to allow running on the latest Windows server VM on GH
- `node-abi` has been updated to allow builds of newer Electron releases
- README status badge has been updated to point at GH actions

cc @kevinmehall


.github/workflows/prebuild.yml
```diff
@@ -0,0 +1,32 @@
+name: prebuild
+
+on:
+  push:
+    branches:
+      - master
```
kevinmehall, line 6: Since this attempts to upload the artifacts to a release (`prebuild -u`), we only want to run this on tags, not on every push to master. It looks like this is where it is failing on the [run on your branch](https://github.com/thegecko/node-usb/runs/1697564231).

We do want something that builds the C++ on at least one version of Node on every branch and pull request. I'm not super familiar with GHA, but maybe that belongs in a separate .yml file vs using conditionals like we did on Travis?

thegecko:
> Since this attempts to upload the artifacts to a release (prebuild -u), we only want to run this on tags, not on every push to master

Ack, that's an oversight on my part, I missed the [tags condition](https://github.com/tessel/node-usb/blob/master/.travis.yml#L65).

> It looks like this is where it is failing on the run on your fork

The failures were due to it trying to push to `tessel/node-usb`, I ran it for [thegecko/node-usb successfully](https://github.com/thegecko/node-usb/actions/runs/483680153) with a temporary url change to prove the workflow.

> maybe that belongs in a separate .yml file vs using conditionals like we did on Travis?

I think I prefer it in a single file (GHA isn't prescriptive AFAIK) so only one file needs to manage the node/electron versions.

I'll update the PR accordingly.

thegecko:
Closing this PR while fixing it...

PR #408, opened 2021-01-30
Don't run portNumbers test on MacOS arm64

thegecko:
This PR suppresses the check for `portNumbers` on MacOS `arm64` architecture as this feature is unsupported.


kevinmehall:
I'm assuming this is arm64 on macOS? I'd expect libusb to work fine on arm64 Linux.

BTW, how do the prebuilds work on the M1 Macs?

thegecko:
> I'm assuming this is arm64 on macOS?
Yes

> I'd expect libusb to work fine on arm64 Linux.
Ok,  I don't have access to that setup. I can restrict to Darwin,  too. 

> BTW, how do the prebuilds work on the M1 Macs?
Well,  it worked :), I think it compiled locally,  though.

thegecko:
Updated to only skip `portNumbers` check on MacOS arm64.

PR #410, opened 2021-01-31
Temporary fix for v11 and v12 Electron builds

karniv00l:
Temporary solution to:

- https://github.com/tessel/node-usb/issues/409
- https://github.com/prebuild/prebuild/issues/273

`node-abi` is a `prebuild` dependency, even the latest released version of `node-abi` returns invalid ABI for `12.0.0-beta.1`. `node-abi` master is working fine but they need to make a release.

---

this produces `usb-v1.6.4-electron-v87-darwin-x64.tar.gz` on my local machine so it doesn't overwrites `v85`

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
   "devDependencies": {
     "coffeescript": "~2.4.1",
     "mocha": "~6.1.4",
-    "node-abi": "^2.19.3",
     "prebuild": "^10.0.1"
   },
   "license": "MIT"
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   "devDependencies": {
     "coffeescript": "~2.4.1",
     "mocha": "~6.1.4",
+    "node-abi": "^2.19.3",
     "prebuild": "^10.0.1"
   },
   "license": "MIT"
--- a/.github/workflows/prebuild.yml
+++ b/.github/workflows/prebuild.yml
@@ -34,7 +34,7 @@ jobs:
       # Supported Node versions: https://nodejs.org/en/about/releases/
       - run: npm run prebuild -- -t 8.0.0 -t 10.0.0 -t 12.0.0 -t 14.0.0 -a ${{ matrix.config.arch }}
       # Supported Electron versions: https://electronjs.org/docs/tutorial/electron-timelines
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -47,7 +47,6 @@
   "devDependencies": {
     "coffeescript": "~2.4.1",
     "mocha": "~6.1.4",
-    "node-abi": "^2.19.3",
```
karniv00l, line 50: @thegecko what was the purpose of that package here? I have removed this since I don't see this used anywhere.

thegecko: I added it to be able to control `node-abi` upgrades without relying on the dependency graph, happy to remove it, however.

thegecko:
So if I understand this correctly, updating the electron 12 version means it no longer overwrites the electron 11 prebuild?

I notice there is a lot of version updates in the `package-lock` file. I'd prefer to see the updates isolated in a separate PR if possible?

karniv00l:
Yes, when we set v12 strictly to `12.0.0-nightly.20201013` `node-abi` returns correct ABI version so we end up with a separate file (https://github.com/lgeiger/node-abi/blob/42932775e72f8886f87b518fadb25d08b5f8c3da/abi_registry.json).

karniv00l:
I have reverted package.json changes

thegecko:
> Yes, when we set v12 strictly to `12.0.0-nightly.20201013` `node-abi` returns correct ABI version so we end up with a separate file (https://github.com/lgeiger/node-abi/blob/42932775e72f8886f87b518fadb25d08b5f8c3da/abi_registry.json).

Perfect!, good fix

thegecko:
> prebuild / prebuild (windows-latest, ia32) (pull_request) Failing after 1m

Looks like GiHub actions can be temperamental, perhaps another push?

karniv00l:
> > prebuild / prebuild (windows-latest, ia32) (pull_request) Failing after 1m
> 
> Looks like GiHub actions can be temperamental, perhaps another push?

✅

thegecko:
Thanks,  good to go I reckon. 

cc @kevinmehall

karniv00l:
Thank you, can we expect a release any time soon? 🙏

thegecko:
> Thank you, can we expect a release any time soon? 🙏

Done!

PR #411, opened 2021-02-02
Define properties as configurable, eliminating errors if library is reloaded

marklangley:
The Jest test runner with --runInBand option enabled runs all tests in a single thread rather than creating a worker pool of child processes. This is essential when testing with node-usb otherwise tests fail randomly as they contend for interfaces and endpoints.

Unfortunately, setting --runInBand causes node-usb to reload in the same process multiple times, once for every test module. And if strict mode is turned on, repeated calls to Object.defineProperty() cause TypeError's to be thrown. This is because the current implementation marks the created properties as non-configurable.

The fix in this pull request is straightforward - simply mark each of the defined properties as configurable, allowing them to update and eliminating the TypeError.

### DIFF
--- a/usb.js
+++ b/usb.js
@@ -58,7 +58,8 @@ Object.defineProperty(usb.Device.prototype, "configDescriptor", {
 			if (e.errno == usb.LIBUSB_ERROR_NOT_FOUND) return null;
 			throw e;
 		}
-	}
+	},
+	configurable: true
 });
 
 Object.defineProperty(usb.Device.prototype, "allConfigDescriptors", {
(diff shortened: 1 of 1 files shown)


thegecko:
I'm not convinced changing the library in this way (and potentially losing guards against type modifications and deletion) is a sensible fix. What is your scenario with jest, are you using it as an external runner?

thegecko:
@marklangley Is this PR still something you want to get in?

marklangley:
@thegecko Thanks for following up! I don't think so. The POC that I was building has since been retired so I have no further need, and I agree that your security considerations are a strong argument against inclusion.

PR #419, opened 2021-05-04
Fix compiler warnings

joelpurra:
Fixing/ignoring compiler warnings (on Ubuntu) to reduce clutter.

- Before: https://github.com/tessel/node-usb/runs/2313113186#step:5:10
- After: https://github.com/tessel/node-usb/runs/2500923044#step:5:10


thegecko:
Thanks, this reduces warnings and I've tested it locally. Will merge and see how the prebuilds get on.

PR #421, opened 2021-05-04
Ignore compiled test file

joelpurra:


thegecko:
What does this change do?

joelpurra:
@thegecko: ignores the compiled test file used in the `valgrind` test.

https://github.com/tessel/node-usb/blob/v1.7.1/package.json#L42

thegecko:
OK, could you add it to the root `.gitignore instead`?

thegecko (approved):
Thanks 👍

PR #443, opened 2021-08-30
v2.0.0 API

thegecko:
Update API version as discussed in https://github.com/tessel/node-usb/issues/413

This version can be installed as follows:

```bash
npm i usb@alpha
```

Please add any feedback to that issue or this PR.

cc @kevinmehall


thegecko:
All functionality is complete and so is the time for feedback. Merging...

PR #445, opened 2021-08-30
fix invalid initial refs

mildsunrise:
it's impossible to close() a device just after open(), you'll always get: `Error: Can't close device with a pending request`

refs should be set initially to 0, since [napi_wrap() returns a weak reference](https://nodejs.org/api/n-api.html#n_api_napi_wrap)

### DIFF
--- a/src/device.cc
+++ b/src/device.cc
@@ -11,7 +11,7 @@
 
 Napi::FunctionReference Device::constructor;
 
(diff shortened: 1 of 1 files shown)


thegecko:
Thanks @mildsunrise , I have confirmed the issue and this patch fixes it.

Issue #124, opened 2016-05-30
Crash with node-usb sending to sleep 


armaseg:
- Operating system: Windows 7 64bits
- Node: 6.1.0

I've been using node-usb for a while and It works really good but it crashs only when I set the pc on sleep mode.

You can reproduce this behavior using this code:

````javascript
var usb = require('usb');
usb.on('attach', function(device) { console.log('usb connected') });
````
and then set the PC on sleep mode.

Regards,

gfcittolin:
I confirm the behavior. Actually, for me, node just exits without any message, but with the exit code `-1073741819`

Using Windows 8.1 and Node 4.2.1

jeanfsal:
I have the same issue on windows 10....anybody knows how to solve this.


I am ready to pay for the fix...

taktran:
We are seeing the same issue too, when [noble](https://github.com/sandeepmistry/noble) is used on windows with a dongle.

rstolwijk:
We are also experiencing this issue on Windows 7 and Windows 8.1 using nw.js v0.15.4
and Node v6.2.2

pablozandona:
I have the same issue on:

- windows 7
- node 6.3.1
- nwjs 0.16.1

bladesling:
@stberger3040 that does indeed fix this issue. Running this in nwjs and no more crashing on sleep. Thank you!

stberger3040:
Thank you @thegecko for taking care of this, I really don't have time. Beware, however that ist is really a quickfix. Devices with an interface path longer than 4096 will stop working with that library. I assumed that this is never the case here (which could be a wrong assumption). The action problem to cause the crash was an integer underflow. `len` was `-1` (which leads to a quite high number when interpreted as `unsigned` ;-)

thegecko:
Thanks @stberger3040

PR #451, opened 2021-10-14
Use libusb with sleep crash fix

### LINKED ISSUE #220: crash electron when the computer sleep then wakeup in windows 10 os
will crash electron when the computer sleep then wakeup in windows 10 os

thegecko:
Update to use fixed libusb to prevent Windows crash on sleep. Fixes #220, #124

Test version available by installing as follows:

```bash
npm i usb@testing
```

Currently untested

cc @jeremywho


jeremywho:
I just tested it. 

No Crash!

Let me know if I can help with any other testing or any other data collection to get this in.

Issue #161, opened 2017-01-05
Plans to switch from forked libusb to upstream libusb?


tony--:
Are there any plans to switch?
Is the lack of hotplug support for windows the primary blocker?
When [hotplug support for windows is added to libusb](https://github.com/libusb/libusb/issues/86) will this module switch to using libusb instead of the fork?

kevinmehall:
Yes, hotplug support is the only difference on the fork, and it would be great to switch to upstream libusb.

PR #453, opened 2021-10-18
Libusb update to v1.0.23

### LINKED ISSUE #229: How to rebuild this package with the latest version of libusb on Windows?
Version 1.0.17 of libusb which is included with this project appears to have an issue when used within a virtual machine.  The getDeviceList function only returns the USB hub object and not the children (usb devices) of that hub.  I downloaded and compiled version 1.0.22 of libusb and the issue appears to be resolved.  (I can get a list of the usb devices in my virtual machine.)  My question is, how can I get the updated version of libusb built into this project?  (I have a bleno based project which uses this package to communicate with a Bluetooth 4.0 USB dongle.)

thegecko:
This PR switches to use the upstream [libusb library](https://github.com/libusb/libusb) at `v1.0.23`.

Based on comments and feedback from @ivannkf, @koush and @tombertrand I wanted to see how broken this is as apparently it's just missing Windows hotplug support.

My testing shows that with this release, hotplugging works on Windows (?!), does anyone know what I'm missing or shall we switch to the upstream release?

If so, this fixes #229 and #161

I've released a test version for people to try and I'd love some feedback as this would be a major change:

```bash
npm i usb@libusb
```

cc @jeremywho, in case this breaks the recent windows sleep crash fix.


kevinmehall:
> My testing shows that with this release, hotplugging works on Windows (?!), does anyone know what I'm missing or shall we switch to the upstream release?

By "hotplug", you mean that the `usb.on('attach', ...)` events work? `libusb_hotplug_register_callback` remains unsupported on Windows in upstream libusb as far as I can tell.

koush:
> Another option that is a bit of a hack: On Windows when the event listener is enabled, you could start a `setInterval` that polls `getDeviceList` and reports events when the list changes.

This is what I did.

prusnak:
Latest stable is `1.0.24`. Is there any reason why not to use that?

thegecko:
> Latest stable is 1.0.24. Is there any reason why not to use that?

To be fair, this started as a PoC and I was wary of the comment from @koush [here](https://github.com/koush/node-usb/commit/23873e8f02898142765357bb391b7904a4519518)

>1.0.24 has a bad commit that breaks config parsing

Happy to move to 1.0.24 if this isn't an issue.

thegecko:
Moving to `libusb` `v1.0.24` caused an issue on Windows with a clashing definition of `ssize_t`. Will stick to `v1.0.23` for now.

PR #470, opened 2022-01-19
Throw if error during initialize()

gniezen:
When there are any errors during `initialize()`, it currently silently fails and returns empty. This can be problematic if there are "access denied" errors when for example udev rules aren't set properly on Linux. Having `initialize()` throw the error would be really helpful in these instances.

### DIFF
--- a/tsc/webusb/webusb-device.ts
+++ b/tsc/webusb/webusb-device.ts
@@ -401,7 +401,7 @@ export class WebUSBDevice implements USBDevice {
             this.serialNumber = await this.getStringDescriptor(this.device.deviceDescriptor.iSerialNumber);
             this.configurations = await this.getConfigurations();
         } catch (error) {
-        	throw new Error(`initialize error: ${error}`);
+            throw new Error(`initialize error: ${error}`);
         } finally {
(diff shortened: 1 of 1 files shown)


tsc/webusb/webusb-device.ts
```diff
@@ -400,6 +400,8 @@ export class WebUSBDevice implements USBDevice {
             this.productName = await this.getStringDescriptor(this.device.deviceDescriptor.iProduct);
             this.serialNumber = await this.getStringDescriptor(this.device.deviceDescriptor.iSerialNumber);
             this.configurations = await this.getConfigurations();
+        } catch (error) {
+        	throw new Error(`initialize error: ${error}`);
```
thegecko, line 404: Please check indent and use spaces not tabs

thegecko:
Thanks for the PR.
As the `initialize()` function doesn't catch any errors, these would already be thrown up the stack, so catching and re-throwing makes no difference.

However, your addition does add useful information as to where the error comes from.

thegecko (changes requested):
Formatting needed

thegecko (approved):

PR #479, opened 2022-02-09
Don't throw when finding endpoints

### LINKED ISSUE #478: getEndpoint fails eagerly
I am in the process of porting our legacy node-webusb to node-usb and found this issue.

In `WebUSBDevice.getEndpoint`, there is a call to `iframe.endpoint(...)`
image

This call fails eagerly instead of letting the loop go through
image

This happens when we try to binding to endpointNumber 5 on a micro:bit v2.

thegecko:
Fixes #478 

cc @pelikhan

### DIFF
--- a/tsc/usb/interface.ts
+++ b/tsc/usb/interface.ts
@@ -161,12 +161,7 @@ export class Interface {
      * The device must be open to use this method.
      * @param addr
      */
-    public endpoint(addr: number): Endpoint {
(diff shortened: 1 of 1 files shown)


pelikhan:
@thegecko when do you plan for your next release? Love the solidification of the usb packages!

pelikhan:
Awesome
________________________________

Issue #495, opened 2022-03-27
LIBUSB_ERROR_NO_DEVICE during startup of Electron app on Windows


sebastianrath:
Hey there, we use `libusb` to detect attach/detach events of USB devices in our Electron app. We recently came across two users that reported the error below.

I'm not sure if this is caused by registering the event handlers, or if this caused by loading the module in the first place. Can anyone shed some light on this?

I found this issue https://github.com/node-usb/node-usb/issues/173 but it only refers to the exception while unplugging a device, which doesn't seem to be the case in our scenario.

![Untitled 2](https://user-images.githubusercontent.com/12844423/160287598-cd4e5463-1b3c-4ee0-91a6-7795f3146c82.png)

mildsunrise:
I think it's a race. a while ago, we replaced the native hotplug patch with an emulation by periodically polling the output of `getDeviceList()` and taking differences:

https://github.com/node-usb/node-usb/blob/54c6e5dfc4cc990d278805b7ed6d32825b17479d/tsc/usb/index.ts#L60-L74

this `deviceAddress` isn't a property, it's a getter that calls `libusb_get_device_address`. it's possible for a device to appear in `getDeviceList` and go away just before calling `libusb_get_device_address`, therefore resulting in a `NO_DEVICE` error. we should handle that error correctly.

YiyuanYin:
I came cross with the same problem,
![image](https://user-images.githubusercontent.com/9218146/165056698-2c594fc1-0466-4c52-8b12-59bab9879a23.png)

thegecko:
Is this still occurring since v2.3.0?

thegecko:
Closing this issue due to inactivity, please create a new issue as required.

Issue #488, opened 2022-03-21
Use native usb device detection on Windows


thegecko:
The current device detection on windows uses a polling mechanism from JavaScript which feels wasteful and dirty.
I plan to try merging the native code from the [node-usb-detection library](https://github.com/MadLittleMods/node-usb-detection) to undertake this in a cleaner way.

Rough plan could be:

- [x] Add native source into node-usb and get it compiling in the CI
- [ ] Wire in the detection start/stop functions into `_enableHotplugEvents` and `_disableHotplugEvents`
- [ ] Find a way of uniquely identifying a device in the detection code in order to:
- [ ] Load the device using libusb and raise the relevant event from node-usb

To this end, I now have the NAPI port of `node-usb-detection` from @Julusian sitting on the [windows-native-detection branch](https://github.com/node-usb/node-usb/tree/windows-native-detection)

Julusian:
I've made a start on this, but am having a hard time translating the windows events into libusb devices.
I am half way through exploring a way to translate, but I am wondering if it is enough for it to report that 'something changed' and use that to trigger a poll.
I don't see a way of getting a `libusb_device` without using `libusb_get_device_list`, so the difference is really whether a diff needs to be done, or if the backend can provide ids to say which device changed (with the hope it wont have disappeared before `libusb_get_device_list` is called). I am starting to feel like it isnt worth the extra effort to do the ids method, and a diff when we know something has changed should be good enough.

thegecko:
>  I am wondering if it is enough for it to report that 'something changed' and use that to trigger a poll.

That would be a lot better than the current polling mechanism and I suspect a changed event may well be enough. Do you want to open a PR for discussion and we can grow from there perhaps?

Issue #481, opened 2022-02-14
Attaching usb doesn't trigger usb.on('attach') event


YiyuanYin:
I have an electron application that use this package to listen to the usb attach event. Sometimes attaching a usb can trigger `usb.on('attach')` event, sometimes it won't.
The code checking whether a usb is attached is by finding the new device which has the different `deviceAddress` from the former one. However, sometimes when a usb is attached, the `deviceAddress` is the same as the old ones in the device list. Therefore, `usb` won't emit `attach` event.
![4DC705B7-442E-432e-BDDA-189A3E5D7FBD](https://user-images.githubusercontent.com/9218146/153823533-5a344d63-2a9a-4616-8fea-1e30474731d0.png)
Have you ever thought about adding some other conditions to decide whether a usb is attached?

thegecko:
> The code checking whether a usb is attached is by finding the new device which has the different deviceAddress from the former one.

The attach/detach events are exposing hotplug events from the [libusb](https://github.com/libusb/libusb) library (except on Windows)

Perhaps you need to raise this issue there?

mildsunrise:
You mentioned "The code checking whether a usb is attached" so I assume you're in Windows.
There's #488 to try to implement actual hotplug detection support on Windows.

mildsunrise:
By the way, I've just realized why this doesn't work... `deviceAddress` is only unique *within the bus the device is connected to*.
To properly identify the device, we need the `busNumber` too. Something like the `deviceId` from WebUSB:

https://github.com/node-usb/node-usb/blob/c42aa55fba09ea65c660d78cdd3779cb29f1d04c/tsc/webusb/index.ts#L347-L353

But I think we can do it even simpler... libusb will never return different `libusb_device` pointers for the same device. And the bindings will never create two `Device` instances for the same `libusb_device`. Therefore we should be able to just say `device1 === device2` to compare both lists of devices.

thegecko:
Closing this issue due to inactivity, please create a new issue as required.

PR #496, opened 2022-03-27
(hopefully) better polling for hotplug

mildsunrise:
While #488 arrives, this should be a safer / more efficient implementation of the polling fallback. Context (and probably fixes):

 - https://github.com/node-usb/node-usb/issues/481#issuecomment-1079999445
 - https://github.com/node-usb/node-usb/issues/495

This relies on the fact that libusb should never create two `libusb_device` for the same logical USB device.
So instead of replicating the comparison ourselves, we can (at least in theory) rely on that fact.


mildsunrise:
I tested this in Linux by hardcoding `hotplugSupported = false` and it works as expected.
still, if you want to merge this, I'd recommend testing on a windows machine first.

thegecko:
Thanks @mildsunrise

Issue #500, opened 2022-04-02
Electron App crash. 


sergeu90:
Hello everyone. I have some strange problem. I use webusb and sometimes my app crash 

Tested on macos and windows. Could you help me please.

```
static async getDeviceByProductId(productId) {
    const customWebUSB = new WebUSB({
      devicesFound: (devices) => devices.find((device) => device.productId === productId),
    });

    return customWebUSB.requestDevice({ filters: [{}] });
  }

  static async checkConnect(productId) 
    const device = await this.getDeviceByProductId(productId);
    await device.open();
    await device.close();
  }
```

```

-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               Electron Helper (Renderer) [99466]
Path:                  /Users/USER/*/Electron.app/Contents/Frameworks/Electron Helper (Renderer).app/Contents/MacOS/Electron Helper (Renderer)
Identifier:            com.github.Electron.helper
Version:               ???
Code Type:             X86-64 (Native)
Parent Process:        Electron [93391]
Responsible:           webstorm [78579]
User ID:               501

Date/Time:             2022-04-02 15:57:58.4093 +0300
OS Version:            macOS 12.2.1 (21D62)
Report Version:        12
Bridge OS Version:     3.0 (14Y910)
Anonymous UUID:        49D68D9F-851C-B63E-5D5B-B3DE0E9F648D

Sleep/Wake UUID:       DDDA2746-9B29-4F7B-BD1C-66B3558D3A9E

Time Awake Since Boot: 670000 seconds
Time Since Wake:       21898 seconds

System Integrity Protection: enabled

Crashed Thread:        0  CrRendererMain  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000091f2a
Exception Codes:       0x0000000000000001, 0x0000000000091f2a
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [99466]

VM Region Info: 0x91f2a is not in any region.  Bytes before following region: 4484931798
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
[...]

PR #505, opened 2022-04-10
Libusb v1.0.26

### LINKED ISSUE #494: v2.2.0 segfaults on Arch Linux
After upgrading to v2.2.0, the following code sometimes segfaults on Arch Linux, killing the entire node process:
```js
const { WebUSB } = require('usb');
const usb = new WebUSB({ allowAllDevices:true })
usb.getDevices().then(console.log).catch(console.error)
```
After reverting to 2.1.3, the same code functions properly.

thegecko:
## Changes

- Upgrade to libusb v1.0.26

## Fixes

- #494
- #500

## Checklist

Have you...

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,6 +11,6 @@
 Have you...
 
 - [ ] Tested the change acts as expected
-- [ ] Checked the change doesn't remove ort change existing functionality
+- [ ] Checked the change doesn't remove or change existing functionality
 - [ ] Considered how the feature impacts the product and tested around it (not just the happy paths)
 - [ ] Where possible, executed the hardware tests on multiple operating systems


mildsunrise (approved):
At some point we should take a closer look at the buildsystem changes, but I wouldn't delay this release more.
In v1.0.26 there's only been two: one for MinGW (https://github.com/libusb/libusb/commit/f9162d36bf7253b3743d2973e385d32166ea0795) and another for some archs like SPARC (https://github.com/libusb/libusb/commit/561dbdacfbcd8f26c81900d1a1a16d2f44093e0e).

Quickly tested in Arch, with Linux 5.17, and the example in #494 no longer crashes.

thegecko:
USB Module
    ✔ should describe basic constants
    ✔ should handle abuse without crashing
    setDebugLevel
      ✔ should throw when passed invalid args
      ✔ should succeed with good args

  getDeviceList
    ✔ should return at least one device

  findByIds
    ✔ should return an array with length > 0

  findBySerialNumber
    ✔ should return a single device 

  Device
    ✔ should have sane properties
    ✔ should have a deviceDescriptor property
    ✔ should have a configDescriptor property
    ✔ should open
    ✔ gets string descriptors
    ✔ supports null string descriptors
    control transfer
      ✔ should OUT transfer when the IN bit is not set
      ✔ should fail when bmRequestType doesn't match buffer / length
      ✔ should IN transfer when the IN bit is set
      ✔ should signal errors
    Interface
      ✔ should have one interface
      ✔ should be the same as the interfaceNo 0
      IN endpoint
        ✔ should be able to get the endpoint
        ✔ should be able to get the endpoint by address
        ✔ should have the IN direction flag
        ✔ should have a descriptor
        ✔ should fail to write
        ✔ should support read
        ✔ times out
        ✔ polls the device (1663ms)
      OUT endpoint
        ✔ should be able to get the endpoint
        ✔ should be able to get the endpoint by address
        ✔ should have the OUT direction flag
        ✔ should support write
        ✔ times out

  WebUSB Module
    ✔ should describe basic constants

  requestDevice
    ✔ should return a device

  getDevices
    ✔ should return one device

  Device properties
    ✔ should have usb version properties
    ✔ should have device version properties
    ✔ should have class properties
    ✔ should have protocol property
    ✔ should have vid/pid properties
    ✔ should have a single configuration
    ✔ should have a configuration property
    ✔ should have a single interface
    ✔ should have a single alternate

  String descriptors
    ✔ gets serialNumber string
    ✔ gets manufacturerName string
    ✔ gets productName string

  Device access
    ✔ is not open
    ✔ is opens and closes

  Configurations
    ✔ selects existing configuration
    ✔ fails to select missing configuration

  Interfaces
    ✔ claims existing interface
    ✔ fails to claim missing interface
    ✔ releases existing interface
    ✔ fails to release missing interface

  Alternates
    ✔ selects existing alternate

  Transfers
    ✔ should control transfer OUT
    ✔ should control transfer IN
    ✔ should transfer OUT
    ✔ should transfer IN


  60 passing (2s)

✨  Done in 2.43s.

Tests pass, merging

Issue #491, opened 2022-03-25
Make the addon context-aware


mildsunrise:
It would be great if we could make node-usb [context aware](https://nodejs.org/api/addons.html#context-aware-addons), i.e. so it can be initialized multiple times from different Node.js instances (node_envs). (Actually the addon incorrectly declares itself as context aware, and that's why users get crashes)

This way it can safely run in Electron and worker threads, probably fixing: #458 #378

This basically requires global variables to be [moved to environment-specific data](https://nodejs.org/api/n-api.html#environment-life-cycle-apis).

> <details><summary>Required changes</summary>
> I've reviewed the code to see the variables we should move, let me know if I'm missing something.
> 
> At least the following static variables in `Device` (I think node-addon-api already saves constructor for us):
> 
> https://github.com/node-usb/node-usb/blob/007e702c3053f4cf649356b4cb3994595a82885e/src/node_usb.h#L70-L71
> 
> And the following hotplug global variables:
> 
> https://github.com/node-usb/node-usb/blob/007e702c3053f4cf649356b4cb3994595a82885e/src/node_usb.cc#L159
> 
> https://github.com/node-usb/node-usb/blob/007e702c3053f4cf649356b4cb3994595a82885e/src/node_usb.cc#L191-L193
> 
> And because each context will use different threads, this as well (or protect with a mutex):
> 
> https://github.com/node-usb/node-usb/blob/007e702c3053f4cf649356b4cb3994595a82885e/src/node_usb.cc#L26
> 
> I think it would be a good idea to use a different libusb context for each Node.js context, so this too:
> 
> https://github.com/node-usb/node-usb/blob/007e702c3053f4cf649356b4cb3994595a82885e/src/node_usb.cc#L15
>
> https://github.com/node-usb/node-usb/blob/007e702c3053f4cf649356b4cb3994595a82885e/src/node_usb.cc#L62
> </details>

As an implication, we also need to implement cleanup (make the addon unloadable) with a [cleanup hook](https://nodejs.org/api/n-api.html#napi_add_env_cleanup_hook). In Windows, we need to tell the thread to exit + join it. Then the tests.

I'm **not** saying I'll have time to do this, just opening this to track the issue (and point people complaining about Electron / workers here) and to see what your thoughts are.

mildsunrise:
@thegecko thanks for taking on this. if I ever have time to work on it myself, I'll hit you up.
Looking more at the code, I think the most complicated part is going to be the cleanup.

[...]

PR #440, opened 2021-08-11
Solves crash on MacOS and Electron +9

danielmain:
Solves crash described here: https://github.com/tessel/node-usb/issues/438

@thegecko  I appreciate a review here. Thank you!

PR #512, opened 2022-04-24
Make addon context aware

thegecko:
## Changes

- Moved global variables into a context structure as outlined in #491

## Fixes

- #491

## Checklist

Have you...

cc @mildsunrise

### DIFF
--- a/src/device.cc
+++ b/src/device.cc
@@ -11,10 +11,7 @@
 
 Napi::FunctionReference Device::constructor;
 
-Device::Device(const Napi::CallbackInfo & info) : Napi::ObjectWrap<Device>(info), device_handle(0), refs_(0)
-#ifndef USE_POLL
-, completionQueue(handleCompletion)
-#endif
+Device::Device(const Napi::CallbackInfo & info) : Napi::ObjectWrap<Device>(info), device_handle(0), refs_(0), completionQueue(handleCompletion)
 {
 	device = info[0].As<Napi::External<libusb_device>>().Data();
 	libusb_ref_device(device);
@@ -193,9 +190,7 @@ Napi::Value Device::Open(const Napi::CallbackInfo& info) {
 	ENTER_METHOD(Device, 0);
 	if (!self->device_handle){
 		CHECK_USB(libusb_open(self->device, &self->device_handle));
-#ifndef USE_POLL
 		completionQueue.start(info.Env());
-#endif
 	}
 	return env.Undefined();
 }
@@ -205,9 +200,7 @@ Napi::Value Device::Close(const Napi::CallbackInfo& info) {
 	if (self->canClose()){
 		libusb_close(self->device_handle);
 		self->device_handle = NULL;
-#ifndef USE_POLL
 		completionQueue.stop();
-#endif        
 	}else{
 		THROW_ERROR("Can't close device with a pending request");
 	}
--- a/src/device.cc
+++ b/src/device.cc
@@ -34,7 +34,7 @@ std::map<libusb_device*, Device*> Device::byPtr;
 
 // Get a V8 instance for a libusb_device: either the existing one from the map,
 // or create a new one and add it to the map.
-Napi::Object Device::get(napi_env env, libusb_device* dev){
+Napi::Object Device::get(Napi::Env env, libusb_device* dev){
 	auto it = byPtr.find(dev);
 	if (it != byPtr.end()) {
 		auto value = it->second->Value();
@@ -87,7 +87,7 @@ Napi::Value Device::Constructor(const Napi::CallbackInfo& info) {
 	return info.This();
 }
 
-Napi::Object Device::cdesc2V8(napi_env env, libusb_config_descriptor * cdesc){
+Napi::Object Device::cdesc2V8(Napi::Env env, libusb_config_descriptor * cdesc){
 	Napi::Object v8cdesc = Napi::Object::New(env);
 
 	STRUCT_TO_V8(v8cdesc, *cdesc, bLength)
--- a/src/device.cc
+++ b/src/device.cc
@@ -410,7 +410,6 @@ Napi::Object Device::Init(Napi::Env env, Napi::Object exports) {
 
 	ModuleData* instanceData = env.GetInstanceData<ModuleData>();
 	instanceData->deviceConstructor = Napi::Persistent(func);
-	instanceData->deviceConstructor.SuppressDestruct();
 
 	return exports;
 }
--- a/src/device.cc
+++ b/src/device.cc
@@ -9,19 +9,24 @@
 
 #define MAX_PORTS 7
 
-Napi::FunctionReference Device::constructor;
-
-Device::Device(const Napi::CallbackInfo & info) : Napi::ObjectWrap<Device>(info), device_handle(0), refs_(0), completionQueue(handleCompletion)
-{
+Device::Device(const Napi::CallbackInfo& info) : Napi::ObjectWrap<Device>(info), env(0), device_handle(0), refs_(0), completionQueue(handleCompletion) {
+	env = info.Env();
 	device = info[0].As<Napi::External<libusb_device>>().Data();
 	libusb_ref_device(device);
+
+	std::map<libusb_device*, Device*> byPtr = env.GetInstanceData<ModuleData>()->byPtr;
 	byPtr[device] = this;
+
 	DEBUG_LOG("Created device %p", this);
 	Constructor(info);
 }
 
-Device::~Device(){
+Device::~Device() {
 	DEBUG_LOG("Freed device %p", this);
+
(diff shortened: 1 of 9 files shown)


thegecko:
Binaries available as [artefacts on the GH action](https://github.com/node-usb/node-usb/actions/runs/2216911767#artifacts)

src/node_usb.cc
```diff
@@ -12,32 +10,68 @@ Napi::Value RefHotplugEvents(const Napi::CallbackInfo& info);
...
+int LIBUSB_CALL hotplug_callback(libusb_context* ctx, libusb_device* device,
+                     libusb_hotplug_event event, void* user_data) {
+	libusb_ref_device(device);
+	ModuleData* instanceData = (ModuleData*)user_data;
+	instanceData->hotplugQueue.post(new HotPlug {device, event, &instanceData->hotplugThis});
+	return 0;
+}
+
+void USBThreadFn(Napi::Env env) {
+	libusb_context* usb_context = env.GetInstanceData<ModuleData>()->usb_context;
```
mildsunrise, line 53: we should be careful not to call any N-API functions from external threads. the thread should receive the ModuleData* directly, IMO

src/node_usb.cc
```diff
@@ -12,32 +10,68 @@ Napi::Value RefHotplugEvents(const Napi::CallbackInfo& info);
...
 
 Napi::Object Init(Napi::Env env, Napi::Object exports) {
 	Napi::HandleScope scope(env);
+	env.SetInstanceData(new ModuleData());
+	ModuleData* instanceData = env.GetInstanceData<ModuleData>();
 
 	initConstants(exports);
 
 	// Initialize libusb. On error, halt initialization.
+	libusb_context* usb_context;
```
mildsunrise, line 65: this should be initialized to `nullptr`, so that (if libusb_init fails) we don't store garbage into `usb_context` in line 67 below (this is so we don't call `libusb_exit` on garbage, see comment below)

src/node_usb.h
```diff
@@ -60,12 +68,23 @@ struct Device: public Napi::ObjectWrap<Device> {
...
+
+	bool hotplugEnabled = 0;
+	libusb_hotplug_callback_handle hotplugHandle;
+	UVQueue<HotPlug*> hotplugQueue;
+	Napi::ObjectReference hotplugThis;
+	std::map<libusb_device*, Device*> byPtr;
+	Napi::FunctionReference deviceConstructor;
+
+    ModuleData(): hotplugQueue(handleHotplug) {
+    }
```
mildsunrise, line 86: In order to clean up when our module is unloaded, we can either register a cleanup hook, but I think we can also add a destructor here that does the cleanup.

Cleanup tasks are (in order):

 1. Stop the hotplug polling thread. I'd do it by adding a `std::atomic<bool>` to `ModuleData`, and checking this boolean in the `while` condition from USBThreadFn. Then in the destructor set it to false, and join the thread. I think `detach` shouldn't be called if we do this.

 2. Destroy the libusb context by calling `libusb_exit` on `usb_context`. (only do this if usb_context != nullptr, and initialize usb_context to nullptr)

There's one detail I still have to think through, which is the garbage collection of the devices we created. It's possible we *do* need a cleanup hook in the end.

mildsunrise (approved):
I've finished reviewing the code. Other than the missing cleanup, there's only some straightforward corrections :)

src/node_usb.cc
```diff
@@ -101,91 +139,65 @@ Napi::Value GetLibusbCapability(const Napi::CallbackInfo& info) {
...
 	Napi::HandleScope scope(env);
+	ModuleData* instanceData = env.GetInstanceData<ModuleData>();
 
-	if (!hotplugEnabled) {
-		hotplugThis.Reset(info.This().As<Napi::Object>(), 1);
-		hotplugThis.SuppressDestruct();
-		CHECK_USB(libusb_hotplug_register_callback(usb_context,
+	if (!instanceData->hotplugEnabled) {
+		instanceData->hotplugThis.Reset(info.This().As<Napi::Object>(), 1);
+		instanceData->hotplugThis.SuppressDestruct();
```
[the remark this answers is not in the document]
thegecko: This was added as a fix here:
https://github.com/node-usb/node-usb/pull/440

Need to check the problem isn't reintroduced

src/node_usb.cc
thegecko: Fixed

src/node_usb.cc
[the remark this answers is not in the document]
thegecko: Removed

src/device.cc
```diff
@@ -403,8 +408,9 @@ Napi::Object Device::Init(Napi::Env env, Napi::Object exports) {
 		});
 	exports.Set("Device", func);
 
-	Device::constructor = Napi::Persistent(func);
-	Device::constructor.SuppressDestruct();
+	ModuleData* instanceData = env.GetInstanceData<ModuleData>();
+	instanceData->deviceConstructor = Napi::Persistent(func);
+	instanceData->deviceConstructor.SuppressDestruct();
```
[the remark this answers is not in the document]
thegecko: Removed

src/node_usb.cc
thegecko: Done

mildsunrise:
In addition to the review, I wanted to list three things that are strictly unrelated to context-aware but would also make the module play 'nicer' with Node.js:

- I just noticed that making `ref` / `unref` no-ops when hotplug isn't enabled can be confusing to the user, since 'enabled' is just an optimization we do behind the scenes. For example a user may do:

  ~~~ js
  usb.refHotplugEvents(); // keep the loop alive
  console.log('waiting for device to be plugged in')
  usb.on('attach', device => {
    // process the device
  });
  ~~~

  which wouldn't work since the call to `refHotplugEvents` is a no-op (because there's no listeners yet)

- Native addons are expected to comply with [async hooks](https://nodejs.org/api/async_hooks.html), a system for tracking stuff that dispatches events. This is done through N-API by creating an [AsyncContext](https://github.com/nodejs/node-addon-api/blob/main/doc/async_context.md) and passing it when calling `MakeCallback`. This isn't a bug or anything, but it would be nice to fix at some point.

src/node_usb.h
thegecko: Ok thanks, let me know your thoughts on the gc and I can look to implement

mildsunrise:
Oh by the way, the comment above isn't necessary or anything, it was more like a list of stuff to improve someday (if I find the time)

src/node_usb.h
mildsunrise: The two tasks above will have to be implemented, that's for sure. But depending on how GC plays out, we may need to implement something more

mildsunrise (approved):
all good, only cleanup missing

thegecko:
> all good, only cleanup missing

I've added cleanup as you suggested. All tests pass, however the main process execution no longer completes and the process hangs.

Was this your concern with the gc?

mildsunrise:
No, it was not. Let me clone the branch and test...

mildsunrise:
Ah it was just a small issue, we forgot to initialize `usb_context` in the constructor, and we also forgot to interrupt libusb before joining the thread. I don't have access to the branch, but you can cherrypick https://github.com/node-usb/node-usb/commit/82a55a2dc9610f5c6a11c970a93696d596da3068 to get the fix.

Just in case, I checked and this is [exactly how Chromium does the events](https://source.chromium.org/chromium/chromium/src/+/main:services/device/usb/usb_context.cc;drc=5666ff4f5077a7e2f72902f3a95f5d553ea0d88d) as well.

mildsunrise:
To simplify debugging in the future, I've also implemented thread naming -- the thread we launch sets its name to `node-usb events` so you can tell it apart from the other ones. It's in https://github.com/node-usb/node-usb/commit/26d8f6b4b389b525a6cf9c1965b38dda707a429d, but I've only tested it in Linux for now.

thegecko:
> we forgot to initialize usb_context in the constructor, and we also forgot to interrupt libusb before joining the thread

Awesome, thank you @mildsunrise this fixes the exit delay and with this addition I think we have a context aware addon!

> To simplify debugging in the future, I've also implemented thread naming

Great, I'll open that on a separate PR due to the new files being added

I've marked this PR as ready and I'll leave it unmerged for another day in case there is any further feedback from others cc @kevinmehall

thegecko:
Grrr, context-aware tests blow up on Windows. I hates Windows

thegecko:
It seems Windows doesn't like the same device being opened from different contexts. Get:

`LIBUSB_ERROR_ACCESS`

mildsunrise:
well... that looks like a limitation of the libusb backend, best we can do is open them an issue (if there isn't one already) and maybe list the limitation somewhere in README. but at least the addon will no longer crash if used from many contexts, which is nice.

(maybe it's not even the libusb backend, but the kernel driver)

PR #514, opened 2022-05-01
label event thread, to make debugging easier

thegecko:
## Changes

-

## Fixes

-

## Checklist

Have you...

### DIFF
--- a/src/thread_name.cc
+++ b/src/thread_name.cc
@@ -1,0 +1,79 @@
+// Definition of a SetThreadName function with an emphasis in "portability",
+// meaning we prefer cautiously failing over creating build errors (or crashes)
+// on exotic platforms. To do this, we only include/link against very
+// few & widely available symbols.
+#include "thread_name.h"
+
+#if defined(__linux__)
+
+    // For Linux use the prctl API directly. prctl symbol
+    // should be available under any libc.
+    #include <sys/prctl.h>
+
+    // Define here to avoid relying on kernel headers being present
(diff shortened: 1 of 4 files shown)


src/thread_name.cc
```diff
@@ -0,0 +1,79 @@
...
+
+    // For Windows, we use the new SetThreadDescription API which
+    // is only available in newish versions. To avoid taking any
+    // risks (and because on certain versions it's the only
+    // option to access the API), we use the dynamic linker
+    #include <windows.h>
+
+    extern "C" typedef HRESULT (WINAPI *SetThreadDescriptionFn)(HANDLE, PCWSTR);
+
+    static auto RetrieveSymbol(const char* objectName) {
```
mildsunrise, line 46:
```suggestion
    static SetThreadDescriptionFn RetrieveSymbol(const char* objectName) {
```

mildsunrise:
tested on Windows and it works correctly as well. it only needed one change, which I have submitted as a suggestion

mildsunrise:
I think we can merge this by now that it's been tested in the 3 OSes (the function does nothing in other OSes)

thegecko:
I've not tested this at all. Is there an easy way to test it on mac/win?

mildsunrise:
yes it's simple to test yourself, just attach a debugger to a Node.js process after `require('usb')`, and you should see a thread named `node-usb events`. if you see it, it's working correctly :)

I don't think we can test it though, but it's fine since it's only a debugging aid

Issue #516, opened 2022-05-23
USB detection loop


adriencarbonaro:
## Bug Description:

When subscribing to `attach` event, the callback triggers in a loop.

## Steps to Reproduce:
1. Launch the following script with `ts-node` or after compiling it :
```
import * as usb from "usb";

usb.usb.on('attach', (device) => {
	console.log(device);
});
```
3. Check the logs. It displays a lot of attached devices, never stopping.

## Additional Information
- Operating System: Windows 10
- Runtime: `node@14.16.1`
- Node USB Version: `usb@2.4.1`

## EDIT

It works with `usb@2.3.1`, and doesn't starting from `usb@2.4.0`

thegecko:
Looks like this was introduced with https://github.com/node-usb/node-usb/pull/496

The devices returned from `getDeviceList()` aren't the same objects each time so `Set.has()` will always be false.

cc @mildsunrise

mildsunrise:
libusb seems to be returning multiple `libusb_device*` referring to the same USB device.
I've reported this upstream.
Maybe I was mistaken, and the "1 USB device = 1 `libusb_device*`" thing isn't a hard guarantee, let's wait for libusb to confirm if this is indeed their bug, or if we should perform additional checks.

mildsunrise:
ah, true. @adriencarbonaro said it works correctly (modulo the 2 attach events) on 2.3.1, so in principle it can't be #496.
could it be because of the context-aware changes?
I'll try to find a windows machine to hopefully reproduce this on, but I'm not sure when that will be

thegecko:
I can confirm it works as expected in `v2.3.1`, so I think it is due to the context-aware rewrite.

You can recreate the issue on any OS by setting `hotplugSupported` to false:

https://github.com/node-usb/node-usb/blob/master/tsc/usb/index.ts#L46

thegecko:
Yup, reverted https://github.com/node-usb/node-usb/commit/12db6e4acd9c4a3a5c388dd0d2aa0b8139dfd6b9 and it works again.

I'll close the other PR but have no idea what's causing the issue, but it's not `libusb`.

mildsunrise:
Okay. I won't close the upstream issue though, since I'd like to be sure the hotplug polyfill is being done correctly.

thegecko:
Fixed in `v2.4.2`

PR #517, opened 2022-05-23
Check device changes by address

thegecko:
## Changes

- Fixed windows hotplug polling to correctly identify devices using address
- Fixes https://github.com/node-usb/node-usb/issues/516

## Fixes

- https://github.com/node-usb/node-usb/issues/516

## Checklist

Have you...


mildsunrise:
The `deviceAddress` is only unique within the USB bus. If we are going to do this, it should be at least based on a `(busNumber, deviceAddress)` tuple.

However, I'd wait for libusb to confirm which checks are needed

PR #519, opened 2022-05-25
fixes in context-aware (#512)

### LINKED ISSUE #516: USB detection loop

mildsunrise:
Aaagh, such a silly error. We were copying the byPtr map instead of obtaining a reference to it.

Refs #512
Fixes #516

### DIFF
--- a/src/device.cc
+++ b/src/device.cc
@@ -14,7 +14,7 @@ Device::Device(const Napi::CallbackInfo& info) : Napi::ObjectWrap<Device>(info),
 	device = info[0].As<Napi::External<libusb_device>>().Data();
 	libusb_ref_device(device);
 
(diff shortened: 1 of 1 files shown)


thegecko:
Confirmed this fixes the issue, thanks @mildsunrise

Issue #521, opened 2022-06-15
Poll Transfer - Start/Stop Poll


noah-moutfitters:
## Bug Description:

Currently not able to make more than one poll transfer without closing the device.

Starting poll transfer has conditional:

```
        if (this.pollTransfers) {
            throw new Error('Polling already active');
        }
```

when no more polls are pending, it resets the state like so:

```
    if (_this.pollPending === 0) {
        _this.pollTransfers = [];
        _this.emit('end');
    }
```

The conditional needs to handle the case of an empty array, since any object in javascript is truthy. Another option is resetting the poll transfers to be undefined.  

I can create a PR for this. 

## Steps to Reproduce:
1. Open a usb device as normal
2. Start a poll transfer
3. Close said poll transfer
4. Wait for the end event to emit
5. Start another poll transfer

thegecko:
A PR would be great!

thegecko:
Does #522 fix you?

noah-moutfitters:
@thegecko yes, thanks!

noah-moutfitters:
@thegecko Do you also want the `pollActive` flag set to false when `pollsPending` equals 0? Seems like that may be desired behavior, it I’m unsure.

thegecko:
Not sure, code was like this before I arrived ;)

thegecko:
> @thegecko Do you also want the pollActive flag set to false when pollsPending equals 0? Seems like that may be desired behavior, but I’m unsure.

I added your suggestion, would you mind checking again?

noah-moutfitters:
@thegecko Yes, thank you!

PR #522, opened 2022-06-16
Check active polling using pollActive

thegecko:
## Changes

- Check active polling using pollActive

## Fixes

- #521 

## Checklist

Have you...


strazzere:
Good catch. This was affecting me over the weekend on the latest deployed to NPM as the `pollTransfers` is never `undefined` but often of length 0.

Was tripping me up a bit until I noticed that the npm deployed code was out of sync with the repo.

thegecko:
I need to do a release

PR #524, opened 2022-07-19
feat: native usb device detection on Windows

Julusian:
## Changes

This is the start of integrating native usb detection.
As there are now two hotplug backends, I split the different implementations into their own files.

I havent found a good way to translate from the path that the detection provides to identifiers that could be matched on a `libusb_device`. And as the only way to obtain a `libusb_device` is through `libusb_get_device_list`, I am unsure if it is worth the extra work.

For this I have added new `attachIds` and `detachIds` events, that report the vendorId and productId of the device that changed.
I think that will by itself be useful for some users. For example when paired with node-hid the `libusb_device` is not useful, but the ids are to know if a scan for new hid devices is appropriate, without needing node-usb to also do a `libusb_get_device_list`

This change is inspired by the `node-usb-detection` code, but uses a newer and simpler api that does not support Windows7 or older. https://docs.microsoft.com/en-us/windows/win32/api/cfgmgr32/nf-cfgmgr32-cm_register_notification.
So far I have only tested it on a Windows10 VM, I will test with more, and check what happens on Windows7

## Fixes

- Closes: #488

## Checklist

Have you...


Julusian:
With a simple test script
```
const lib = require('.')

console.log('can hotplug', lib.usb._supportsHotplugEvents())

lib.usb.on('attach', function(...args) { 
    console.log('attach', ...args)
 });
lib.usb.on('detach', function(...args) { 
    console.log('detach', ...args)
 });

 
lib.usb.on('attachIds', function(...args) { 
    console.log('attachIds', ...args)
 });
lib.usb.on('detachIds', function(...args) { 
    console.log('detachIds', ...args)
 });
```

I get the output of
```
detachIds 4057 96
attachIds 4057 96
```

thegecko:
Great work, thanks for this. I'll try it out a bit later.

Julusian:
I have checked and it works on windows8.1.
It does not work on windows7, but as latest nodejs doesnt run on windows7 this may not be an issue? I think 13.5 is the latest officially supported, electron I believe works in later versions. It does result in failing to load node-usb at all, its not a graceful degredation.

src/hotplug/libusb.cc
```diff
@@ -0,0 +1,92 @@
+#include "hotplug.h"
```
thegecko, line 1: nit: this file has tab and space indents

src/hotplug/hotplug.h
```diff
@@ -0,0 +1,22 @@
+#ifndef _USB_HOTPLUG_H
```
thegecko, line 1: nit: this file has tab indents, please use spaces

src/hotplug/windows.cc
```diff
@@ -0,0 +1,161 @@
+#include "hotplug.h"
```
thegecko, line 1: nit: this file has tab and space indents

src/node_usb.cc
```diff
@@ -1,47 +1,18 @@
 #include "node_usb.h"
```
thegecko, line 1: nit: this file has tab and space indents

thegecko (changes requested):

thegecko:
Awesome, thanks for this PR @Julusian 

I'm finding it quite hard to follow the different `hotplugSupported` types implemented - it has changed a lot of the code (event signatures, polling once vs continuous, etc.).

As an initial pass, I'd prefer to see a simpler "something changed" from the backend code and we re-list the devices from TypeScript. This should then allow us to remove polling entirely.

We could then discuss layering in some of these other changes as follow-ups?

Keen to know your thoughts and what others think (@mildsunrise)?

Julusian:
for the tabs vs spaces, it looks like most files are using tabs, only a couple are using spaces. You asked for spaces, so my new files are spaces, any changes to the others I have made sure are consistent with their existing formatting

Julusian:
> This should then allow us to remove polling entirely.

It might be worth keeping the poll logic around, for the other less common platforms that dont have libusb hotplug support. I dont have a compelling argument for it really, as those platforms are not tier 1 targets for nodejs itself.

> As an initial pass, I'd prefer to see a simpler "something changed" from the backend code and we re-list the devices from TypeScript.

That is what I was trying to achieve here, but I shall do another pass and see if any of my changes can be rewritten/simplified.   
The `attachIds` and `detachIds` are pretty much a 'something changed', but with details that can make them useful to users.

I have stripped out as much of the typescript changes as I can without breaking the new functionality. This makes the `attachIds` and `detachIds` events be 'internal' for now. I would really like them to be public, but that can definitely be a follow up PR. It is mostly intended a simplification/performance optimisation for some use cases, such as pairing with node-hid.

Julusian:
Also, if anyone has a suggestion for better names for `attachIds` and `detachIds`, I think they could do with something better. 
They are written to have the `idVendor` and `idProduct` of the affected device as parameters

mildsunrise:
I'm only following this lightly, so apologies if what I ask doesn't make sense or has previously been discussed. VID/PID don't uniquely identify a device, to do that we would need two things: the **bus number**, and the **device address** within the bus. Does the Windows API offer that info?

If it does, it would allow us to reliably match events to the exact `libusb_device` responsible for them.

mildsunrise:
Hmm, looking quickly through the API it seems the properties are there?
https://docs.microsoft.com/en-us/windows-hardware/drivers/install/devpkey-device-busnumber
https://docs.microsoft.com/en-us/windows-hardware/drivers/install/devpkey-device-address
But they're at "device instance" level, not at "device interface". I'm not sure if you can query them on interfaces as well, or if you need to obtain a handle to the device instance. People seem to do that by calling `CM_Get_Device_Interface_Property()` with `DEVPKEY_Device_InstanceId`, then `CM_Locate_DevNode` to get the handle. Have you tried that?

(Also when dealing with hotplug detection let's remember to handle any races correctly, such as devices no longer existing when we try to look them up at libusb, if we decide to do that.)

Julusian:
@mildsunrise 

I was trying to get it to report the bus number, and the device address, but am now feeling like there isnt any point in doing that.
The only way I can see to get a `libusb_device` is through `libusb_get_device_list()`, so doing a targetted diff is going to be a lot more complicated and less performant than the simple diff everything that this PR currently does. (especially when you consider that we will need to cache the devices to emit for a detach event, but that cant just be the result of the last scan as 2 detach events could be emitted very close together).

I was using those methods you suggest for getting the data, but it was getting rather complicated and involved copying and slightly modifying over 100 lines of code from libusb. They try other properties before Address, so if we need it to match up, we will need to copy their method of determinig the address https://github.com/libusb/libusb/blob/master/libusb/os/windows_winusb.c#L1226. As such I dont think I could do enough testing to satisfy to myself that the ids it produced would be consistently correct. 

>  VID/PID don't uniquely identify a device

It doesnt, but for my intended use that is fine, it allows for usage a bit more like how https://github.com/MadLittleMods/node-usb-detection#usbdetectoneventname-callback can be used.
For node-usb, these events can be treated internally as 'something changed', with the parameters being irrelevant.
But I can see value in these for users. Is it really a good idea to do a full poll and diff for every usb change when you are watching for a certain vid/pid combo? If instead of listening to attach you listen to attachIds, you can do some analysis of 'does this attach/detach look interesting', and do a diff only if something interesting has changed.

For example, node-hid does not use node-usb, and only uses libusb internally on linux. To open a device in node-hid you need the path, which on windows will look something like `'\\?\hid#vid_05f3&pid_0405&mi_00#7&56cf813&0&0000#{4d1e55b2-f16f-11cf-88cb-001111000030}'`. So unless we start reporting this path (which I dont think libusb provides) the parameters arent useful for knowing exactly which device changed. But it does let node-hid users decide whether it is worth doing a `hid_enumerate()`.

And yes, this is possible to do with the devices emitted by attach/detach, but its rather wasteful for node-usb to do a `libusb_get_device_list()` to get the devices that change, for users to then only look at the vid and pid of those devices. These events can be made to allow them to bypass that extra `libusb_get_device_list()` call if they want to (in a follow up PR)

mildsunrise:
Very fair point. Given that libusb doesn't expose any way to get access to a particular libusb_device, these events are only useful as indicators that a re-enumeration + diff should be done. Let's leave this PR as it is then :) It's already better than what we had

README.md
```diff
@@ -547,6 +547,40 @@ Restore (re-reference) the hotplug events unreferenced by `unrefHotplugEvents()`
...
+
+There is no direct equivalent to these methods. This is handled automatically for you when you add and remove event listeners.
+
+You may find `usb.unrefHotplugEvents()` useful as it is intended to help with exit conditions.
+
+## usbDetect.find()
+
+You can instead use `usb.getDeviceList()`. Be aware that this will do an enumeration to find all of the devices, and not look at a cache of the known devices. If you call it too often it may have a performance impact.
+
+The api does not allow for filtering by vid and pid, but you can do this yourself by doing a `usb.getDeviceList().filter((dev) => dev.deviceDescriptor.idVendor === 0x12 && dev.deviceDescriptor.idProduct === 0x34)`
```
thegecko, line 564: This is incorrect, you can use `findByIds` https://github.com/node-usb/node-usb#find-legacy-device-by-vidpid

README.md
```diff
@@ -547,6 +547,40 @@ Restore (re-reference) the hotplug events unreferenced by `unrefHotplugEvents()`
...
+
+## usbDetect.on('add', function(device) { ... })
+
+These should be changed to `usb.on('attach', function(device) { ... })`.
+
+There is no equivalent to filter based on the vid or pid, instead you should do a check inside the callback you provide.  
+The contents of the device object has also changed
```
thegecko, line 571:
```suggestion
The contents of the device object has also changed.
```

README.md
```diff
@@ -547,6 +547,40 @@ Restore (re-reference) the hotplug events unreferenced by `unrefHotplugEvents()`
...
+
+## usbDetect.on('remove', function(device) { ... })
+
+These should be changed to `usb.on('detach', function(device) { ... })`.
+
+There is no equivalent to filter based on the vid or pid, instead you should do a check inside the callback you provide.  
+The contents of the device object has also changed
```
thegecko, line 578:
```suggestion
The contents of the device object has also changed.
```

src/device.cc
```diff
@@ -221,9 +221,9 @@ struct Req: Napi::AsyncWorker {
 		device->ref();
 	}
 
-    void OnOK() override {
+	void OnOK() override {
```
thegecko, line 224: Change has introduced tabs instead of spaces :/

src/hotplug/windows.cc
```diff
@@ -0,0 +1,161 @@
...
+    
+    auto string = std::wstring(buf);
+    std::transform(string.begin(), string.end(), string.begin(), std::towupper);
+
+    wchar_t* temp = new wchar_t[5];
+    temp[4] = L'\0';
+
+    const wchar_t *vidStr = wcsstr(string.data(), VID_TAG);
+    const wchar_t *pidStr = wcsstr(string.data(), PID_TAG);
+
```
thegecko, line 43: nit: double line space

src/node_usb.h
```diff
@@ -76,13 +73,13 @@ struct ModuleData {
...
 	bool hotplugEnabled = 0;
-	libusb_hotplug_callback_handle hotplugHandle;
+	std::unique_ptr<HotPlugManager> hotplugManager;
 	UVQueue<HotPlug*> hotplugQueue;
 	Napi::ObjectReference hotplugThis;
 	std::map<libusb_device*, Device*> byPtr;
 	Napi::FunctionReference deviceConstructor;
 
-    ModuleData(libusb_context* usb_context);
+	ModuleData(libusb_context* usb_context);
```
thegecko, line 82: tabs in this file

thegecko (changes requested):
Some minor change requests - happy to keep tab/space formatting for another PR, though. I still need to test this on all the target hardware...

thegecko:
Tested on macOS 👍

thegecko:
Tested on Win64 👍

README.md
```diff
@@ -547,6 +547,40 @@ Restore (re-reference) the hotplug events unreferenced by `unrefHotplugEvents()`
...
+
+There is no direct equivalent to these methods. This is handled automatically for you when you add and remove event listeners.
+
+You may find `usb.unrefHotplugEvents()` useful as it is intended to help with exit conditions.
+
+## usbDetect.find()
+
+You can instead use `usb.getDeviceList()`. Be aware that this will do an enumeration to find all of the devices, and not look at a cache of the known devices. If you call it too often it may have a performance impact.
+
+The api does not allow for filtering by vid and pid, but you can do this yourself by doing a `usb.getDeviceList().filter((dev) => dev.deviceDescriptor.idVendor === 0x12 && dev.deviceDescriptor.idProduct === 0x34)`
```
thegecko, line 564:
```suggestion
To find a specific device by vid and pid, call `usb.findByIds`. e.g. `usb.findByIds(0x12, 0x34)`.
```

thegecko (approved):

src/hotplug/windows.cc
thegecko, line 43:
```suggestion

```

thegecko (approved):

thegecko:
Great, merging once the prebuilds finish

PR #525, opened 2022-07-19
chore: remove unused header

Julusian:
## Changes

- `node_buffer.h` is not a node-api header, and its contents do not follow the node-api stability guarantees https://nodejs.org/api/n-api.html#implications-of-abi-stability.
  After removing the header, it still compiles so it appears that nothing from it was being used

## Fixes

-

## Checklist

Have you...

### DIFF
--- a/src/node_usb.h
+++ b/src/node_usb.h
@@ -10,9 +10,9 @@
 #endif
 
 #include <thread>
+#include <atomic>
 #include <libusb.h>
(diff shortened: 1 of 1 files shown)


thegecko:
Thanks @Julusian 
I notice an `atomic` reference also snuck in, what does this add?

thegecko (approved):
LGTM

PR #531, opened 2022-08-13
Fix initial windows device list

### LINKED ISSUE #529: Initial detaching of a device doesn't trigger detach event.
## Bug Description:

Starting the app with the device plugged in, then physically unplugging it doesn't trigger `usb.on('detach', (device) => {`. If I unplug, then plug, then unplug the device again, then my detach event finally fires. Also, if I start the app with the device unplugged, it works on the first detach. 

It seems like a detach event can't happen before an attach event, or something. I noticed that a bunch of attach events are fired when *detaching* initially. But none have my device id, so I don't know what's going on. Please advise. Thank you.

## Steps to Reproduce:
1. Plug in your device
2. Detach it (only attach events are fired)
3. Plug it back in (attach event is fired with that device id)
4. Detach it again, this time you will see that device id being detached.

## Additional Information
- Operating System: Windows 10 64 bit.
- Runtime node v16.16.0
- Node USB Version (`npm ls usb`): 2.5.0

thegecko:
## Changes

- Set initial device list for windows polling

## Fixes

- Fixes #529

## Checklist

Have you...

cc @mildsunrise  @Newbrict @Julusian

### DIFF
--- a/src/hotplug/libusb.cc
+++ b/src/hotplug/libusb.cc
@@ -61,12 +61,11 @@ void handleHotplug(HotPlug* info) {
     Napi::Object v8dev = Device::get(env, dev);
     libusb_unref_device(dev);
 
-    Napi::Value v8Vid = env.Null();
-    Napi::Value v8Pid = env.Null();
+    Napi::Object v8VidPid = Object::New(env);
     auto deviceDescriptor = v8dev.Get("deviceDescriptor");
     if (deviceDescriptor.IsObject()) {
-        v8Vid = deviceDescriptor.As<Napi::Object>().Get("idVendor");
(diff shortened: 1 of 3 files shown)


Newbrict:
I will try to figure out how to pull this into my package.json, and if I'm able to, I'll test it. Haha.

Newbrict:
> I will try to figure out how to pull this into my package.json, and if I'm able to, I'll test it. Haha.

Any tips appreciated. I'm still lost.

thegecko:
Clone it recursively, switch branches, yarn and yarn link

Newbrict:
Sorry for the delay, I've been working on something else, but I will get back to this ASAP.

thegecko:
Refactored the code in this area for readability and confirmed this fixes initial device listing on Windows

PR #536, opened 2022-08-29
Trim npm package

### LINKED ISSUE #534: It is recommended to exclude the source code and map files when publishing the NPM package, and only publish the final product

thegecko:
## Changes

- Removes config files from npm package
- Removes typescript source from npm package

While I'm keen to strip the package down and remove the unneeded config files, removal of the c++ source will mean the native module will no longer build/rebuild on unsupported systems and we would rely entirely on the shipped prebuild binaries.

## Fixes

- Fixes #534 

## Checklist

Have you...

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -13,7 +13,4 @@ test/*.js
 .gitmodules
 tsconfig.json
 typedoc.json
-libusb
-libusb_config
-src
 tsc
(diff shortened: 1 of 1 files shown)


thegecko:
Restricted to only remove typescript source from the package

Issue #540, opened 2022-09-07
Native code crashing


gcampbell-msft:
## Bug Description:
I recently upgraded to node-usb 2.5.*, and I started noticing that the native code was crashing my program when plugging in or unplugging hardware. I was able to debug it with VS and found it hit a call stack inside of PNP (plug and play) in native code. 

## Steps to Reproduce:
1. Create a program where `usb.on("attach")` and `usb.on("detach")` is used.
2. Run the program and disconnect or connect a board.
3. Notice the failure.

Additional info: 
I downgraded to 2.4.3 and the issue went away.
Exception:
![image](https://user-images.githubusercontent.com/86264750/188963177-960d0dfb-6510-4bb1-93ae-8e4aedab0859.png)

Call stack:
![image](https://user-images.githubusercontent.com/86264750/188962291-956f6274-711a-4986-b8e5-49ae762c0774.png)

## Additional Information
- Operating System: Windows 10
- Runtime (`node -v` or `electron -v`): 16.13.1, node.
- Node USB Version (`npm ls usb`): 2.5.1

thegecko:
Versions suggest this has been introduced with the Windows USB detection code in https://github.com/node-usb/node-usb/pull/524

mildsunrise:
the fact that it's when delivering a PNP event also strongly suggests this. @Julusian have you seen this previously in node-usb-detection?

Julusian:
No, not something I've seen there, but the code here was written differently and with a newer api.

I have a theory, with a hopeful fix #541

thegecko:
The fix from @Julusian has been released using the `alpha` tag and can be installed using:

```bash
npm i usb@alpha
```

Could you test this fixes your issue?

gcampbell-msft:
@thegecko Yes, I'll try and test today. Thanks!

gcampbell-msft:
I just tested and it did not fix the issue. 

Below are two more images of the callstack and the exception.
![image](https://user-images.githubusercontent.com/86264750/189670271-aea7e58c-c393-4422-ae47-0500d42b6ba6.png)
![image](https://user-images.githubusercontent.com/86264750/189670326-13d16c5f-8d63-4bca-b53b-960bc280e440.png)

Julusian:
@gcampbell-msft any chance of a reproduction script?
I don't think this is going to be an easy one to figure out, as that stack trace is completely inside windows code, and there isnt really much setup/teardown of the events for things to be done wrong..

[...]

PR #541, opened 2022-09-08
fix: windows hotplug cmNotifyFilter access violation #540

Julusian:
## Changes

- Ensure cmNotifyFilter is kept alive for the duration of the hotplug detetion

## Fixes

- #540

## Checklist

Have you...

### DIFF
--- a/src/hotplug/windows.cc
+++ b/src/hotplug/windows.cc
@@ -78,6 +78,7 @@ DWORD MyCMInterfaceNotification(HCMNOTIFICATION hNotify, PVOID Context, CM_NOTIF
 
 class HotPlugManagerWindows : public HotPlugManager
 {
+public:
     HotPlugManagerWindows()
     {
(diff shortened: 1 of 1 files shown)


Julusian:
I don't have access to windows for a few days to test this, or check it compiles.

---- 

I suspect that the fault is that the pointer `&cmNotifyFilter` is being passed into `CM_Register_Notification()`, then `cmNotifyFilter` is being freed almost immediately with the function return.
This pattern was based on node-usb-detection, but in that the function doesnt return until the callback has been unregistered.

So hopefully making sure that `cmNotifyFilter` will live for the whole duration the callback is registered will resolve this

thegecko:
I'll create a pre-release for testing.  Looks like one of the prebuilds failed,  though.

Julusian:
@thegecko fixed!

thegecko (approved):

PR #543, opened 2022-09-09
Update bindings to include Endpoints and update node-gyp

strazzere:
## Changes
- Update node-gyp to a more recent version, this is required for installing on newer version of Windows. This allows use to use python3+ and MSVS 2022.
- Extending the bindings to include `Endpoint`, `InEndpoint` and `OutEndpoint`. This is required to properly use these types in ts.

## Checklist

I have tested this across Windows 10, 11 and on my Linux ubuntu box locally. Annoyingly `npm install` has broken usage with submodules for quiet a while. So all of these had been doing using `npm link` locally after properly updating the submodule manually.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
   "dependencies": {
     "@types/w3c-web-usb": "1.0.6",
     "node-addon-api": "^4.2.0",
-    "node-gyp-build": "^4.3.0"
+    "node-gyp-build": "^4.5.0"
   },
   "devDependencies": {
     "@types/node": "^14.14.41",
@@ -63,7 +63,7 @@
     "coffeescript": "^2.5.1",
     "eslint": "^7.29.0",
     "mocha": "^9.1.3",
-    "node-gyp": "^7.1.2",
+    "node-gyp": "^9.1.0",
     "prebuildify": "^4.2.1",
     "prebuildify-ci": "^1.0.5",
     "prebuildify-cross": "^4.0.1",
--- a/tsc/usb/bindings.ts
+++ b/tsc/usb/bindings.ts
@@ -3,7 +3,7 @@
 //  Rob Moran <https://github.com/thegecko>
 
 import { join } from 'path';
-import type { DeviceDescriptor, ConfigDescriptor, BosDescriptor } from './descriptors';
+import type { EndpointDescriptor, DeviceDescriptor, ConfigDescriptor, BosDescriptor } from './descriptors';
 
 /* eslint-disable @typescript-eslint/no-var-requires */
 const usb = require('node-gyp-build')(join(__dirname, '..', '..'));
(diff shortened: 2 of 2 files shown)


thegecko:
The binding types already exist here:

https://github.com/node-usb/node-usb/blob/master/tsc/usb/endpoint.ts

What problem were you trying to solve with the additional types?

thegecko:
Prebuilds have failed, looks like python 3.6 is now required to build

strazzere:
I will fix the prebuilds and dig deeper.

I was under the impression when utilizing the library that these types where not exposed.

The main types file is `dist/index.d.ts` with directly exports the `dist/usb/index.d.ts`, however niether of these files expose the other types, etc `dist/usb/endpoint.d.ts`. Originally I figured by exposing them via `bindings.ts` would be a good solution, though on relooking at the problem, it might be cleaner and easier to expose them via something like this;

```
diff --git a/tsc/index.ts b/tsc/index.ts
index 8d94fd7..b8f1bf6 100644
--- a/tsc/index.ts
+++ b/tsc/index.ts
@@ -2,6 +2,8 @@ import { promisify } from 'util';
 import { WebUSB, getWebUsb } from './webusb';
 import { WebUSBDevice } from './webusb/webusb-device';
 import * as usb from './usb';
+import * as endpoint from './usb/endpoint';
+import * as descriptors from './usb/descriptors';

 const webusb = new WebUSB();
 const getDeviceList = usb.getDeviceList;
@@ -57,6 +59,9 @@ export {
     // Core usb object for quick access
     usb,

+    descriptors,
+    endpoint,
+
     // Convenience methods
     useUsbDkBackend,
     getDeviceList,
```

Do you have a preference? Or maybe I'm completely missing the boat on how to interact with these types and they are already exposed, though I'm not finding anyway to import these types directly for using in my own typescript projects without doing one of the above.

If you have a preference, let me know and I can clean up the PR and try to fix the prebuilds as well.

Thanks!

thegecko:
> I was under the impression when utilizing the library that these types where not exposed.

The types correctly come through if you use anything of that type:

Screenshot 2022-09-10 at 21 16 28

And if you want to declare a variable with one of these types you can dig deep into the package with:

```typescript
import { InEndpoint } from 'usb/dist/usb/endpoint';
```

e.g.

Screenshot 2022-09-10 at 21 18 29

So technically these are exposed. However, they aren't exposed at the top level. I'll open a separate PR to discuss bringing all types to the top level. Can you make this PR about the node-gyp upgrade only?

strazzere:
Hmm odd. I could have sworn that type of import wasn't working for me. I'll
have to recheck when I'm back at a computer.

Thanks!

thegecko:
All types exposed in top level in #544

thegecko:
Could you update this PR to focus just on the node-gyp upgrade?

PR #544, opened 2022-09-10
Export all usb types at top level

thegecko:
## Changes

As discussed in #543 this PR exposes all usb typescript types at the top level.

## Checklist

Have you...


thegecko:
No concerns raised round this and I don't think it does any harm.

Issue #566, opened 2023-01-22
getDevices excludes a device for which transferIn request is awaiting


nisargjhaveri:
## Bug Description:
`getDevices` call using the WebUsb interface does not include a device which was earlier available and has an active transferIn request.

## Steps to Reproduce:
1. List devices using `getDevices`.
2. Filter the relevant device, open and claimInterface
3. Start a transferIn request, don't await.
4. Try to call `getDevices` again.
5. Notice that the result of the `getDevices` does not include the device.

Here is a snippet I was able to reproduce this with.

```ts
import { WebUSB } from 'usb';

const WebUsbDeviceFilter: USBDeviceFilter = {
    classCode: 0xFF,
    subclassCode: 0x42,
    protocolCode: 1,
};

const usb = new WebUSB({
    "allowAllDevices": true
});

(async () => {
    let devices: USBDevice[];
    devices = await usb.getDevices();

    console.log("Initial device length:", devices.length);

    let device = devices
        .filter(d => {
            return d.configuration?.interfaces.some(iface => {
                return iface.alternate.interfaceClass === WebUsbDeviceFilter.classCode
                    && iface.alternate.interfaceSubclass === WebUsbDeviceFilter.subclassCode
                    && iface.alternate.interfaceProtocol === WebUsbDeviceFilter.protocolCode
            });
        })[0];

    await device.open();

    let inEndpointNumber, outEndpointNumber;
    for (const configuration of device.configurations) {
        for (const interface_ of configuration.interfaces) {
            for (const alternate of interface_.alternates) {
                if (alternate.interfaceClass === WebUsbDeviceFilter.classCode &&
                    alternate.interfaceSubclass === WebUsbDeviceFilter.subclassCode &&
                    alternate.interfaceProtocol === WebUsbDeviceFilter.protocolCode) {
                    if (device.configuration?.configurationValue !== configuration.configurationValue) {
                        await device.selectConfiguration(configuration.configurationValue);
                    }

                    if (!interface_.claimed) {
                        await device.claimInterface(interface_.interfaceNumber);
                    }

                    if (interface_.alternate.alternateSetting !== alternate.alternateSetting) {
                        await device.selectAlternateInterface(interface_.interfaceNumber, alternate.alternateSetting);
                    }

                    for (const endpoint of alternate.endpoints) {
                        switch (endpoint.direction) {
                            case 'in':
                                inEndpointNumber = endpoint.endpointNumber;
                                break;
                            case 'out':
[...]

PR #567, opened 2023-01-23
Return the same instance of USBDevice from separate getDevices calls

nisargjhaveri:
## Changes

- Currently each call to `getDevices` on the webusb interface returns new instances of `USBDevice` even for the same connected device. This causes multiple different issues.
- Instead of creating a new instance every time, cache the `WebUSBDevice` instances corresponding to the underlying `Device` instances and return the cached instance to ensure we always return the same `USBDevice` instance for the same device.
- As far as I could see, this is the behaviour in web, and should be the intended behaviour for us as well.

## Fixes

- #566

## Checklist

Have you...

### DIFF
--- a/tsc/webusb/index.ts
+++ b/tsc/webusb/index.ts
@@ -246,13 +246,13 @@ export class WebUSB implements USB {
 
         const refreshedKnownDevices = new Map<usb.Device, WebUSBDevice>();
 
-        await Promise.all(devices.map(async (device) => {
+       for (const device of devices) {
             const webDevice = await this.getWebDevice(device);
 
             if (webDevice) {
                 refreshedKnownDevices.set(device, webDevice);
             }
-        }));
+        }
 
         // Refresh knownDevices to remove old devices from the map
         this.knownDevices = refreshedKnownDevices;
--- a/tsc/webusb/index.ts
+++ b/tsc/webusb/index.ts
@@ -245,44 +245,38 @@ export class WebUSB implements USB {
         devices = this.preFilterDevices(devices, preFilters);
 
         const refreshedKnownDevices = new Map<usb.Device, WebUSBDevice>();
-        const webDevices: USBDevice[] = [];
 
-        for (const device of devices) {
-            if (this.options.deviceTimeout) {
-                device.timeout = this.options.deviceTimeout;
-            }
-
+        await Promise.all(devices.map(async (device) => {
             const webDevice = await this.getWebDevice(device);
 
             if (webDevice) {
-                webDevices.push(webDevice);
                 refreshedKnownDevices.set(device, webDevice);
             }
-        }
+        }));
 
         // Refresh knownDevices to remove old devices from the map
         this.knownDevices = refreshedKnownDevices;
 
-        return webDevices;
+        return [...refreshedKnownDevices.values()];
     }
 
     // Get a WebUSBDevice corresponding to underlying device.
     // Returns undefined the device was not found and could not be created.
-    private async getWebDevice(device: usb.Device): Promise<WebUSBDevice|undefined> {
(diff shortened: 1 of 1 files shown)


thegecko:
Thanks, your suggested change makes sense. I'll test it out.

tsc/webusb/index.ts
```diff
@@ -257,32 +244,45 @@ export class WebUSB implements USB {
...
+
+        // Refresh knownDevices to remove old devices from the map
+        this.knownDevices = refreshedKnownDevices;
 
+        return webDevices;
+    }
+
+    // Get a WebUSBDevice corresponding to underlying device.
+    // Returns undefined the device was not found and could not be created.
+    private async getWebDevice(device: usb.Device): Promise<WebUSBDevice|undefined> {
```
thegecko, line 271:
```suggestion
    private async getWebDevice(device: usb.Device): Promise<WebUSBDevice | undefined> {
```

thegecko (changes requested):
All hardware tests pass and the CI hasn't broken!

There's a few suggestions, but I also think the `isAllowedDevice()` check and the `device.timeout` addition can be refactored into your new `getWebDevice()` function.

`loadDevices()` can then be simplified further to:

```typescript
    private async loadDevices(preFilters?: USBDeviceFilter[]): Promise<USBDevice[]> {
        let devices = usb.getDeviceList();

        // Pre-filter devices
        devices = this.preFilterDevices(devices, preFilters);
        const webDevices = await Promise.all(devices.map(device => this.getWebDevice(device)));
        return webDevices.filter(device => device) as USBDevice[];
    }

```

nisargjhaveri:
Thanks for your suggestions, makes sense! I've made some changes to simplify `getWebDevice` and `loadDevices`. In the current structure, `isAllowedDevice` cannot be moved to `getWebDevice`, since `requestDevice` needs a list of unfiltered devices to work. Have a look?

tsc/webusb/index.ts
```diff
@@ -257,32 +244,39 @@ export class WebUSB implements USB {
...
-        const webDevices: USBDevice[] = [];
+        const refreshedKnownDevices = new Map<usb.Device, WebUSBDevice>();
+
+        await Promise.all(devices.map(async (device) => {
+            const webDevice = await this.getWebDevice(device);
+
+            if (webDevice) {
+                refreshedKnownDevices.set(device, webDevice);
+            }
+        }));
```
thegecko, line 255:
```suggestion
        devices.forEach(async device => {
            const webDevice = await this.getWebDevice(device);

            if (webDevice) {
                refreshedKnownDevices.set(device, webDevice);
            }
        });
```

tsc/webusb/index.ts
```diff
@@ -257,32 +244,39 @@ export class WebUSB implements USB {
...
+
+        // Refresh knownDevices to remove old devices from the map
+        this.knownDevices = refreshedKnownDevices;
+
+        return [...refreshedKnownDevices.values()];
```
thegecko, line 260:
```suggestion
        return [...this.knownDevices.values()];
```

thegecko (changes requested):
Some minor suggestions

tsc/webusb/index.ts
nisargjhaveri: We'd want to await for all the `getWebDevice` to complete before going ahead, no? Using `forEach` might not do that.

thegecko: Your right, it need to be a normal loop...

thegecko: Suggestion updated

nisargjhaveri: Committed.

thegecko (approved):
Thanks for the fixes!

PR #605, opened 2023-04-29
bump compilation to ES2019

### LINKED ISSUE #604: compile source to something more modern
The generated source code and stack trace on errors are horrific to debug and do hot fixes inside of node_modules/usb/.

[looking](https://www.npmjs.com/package/usb?activeTab=code) at the source code in the dist folder is old-ish. I guess more modern code would ship less, work faster, and be easier to debug.

could you release some more modern ESM, es2022 variant or something else?
make some breaking changes

mildsunrise:
## Changes

according to node.green, Node v10.8.0 supports ES2019, so switch to that instead of the current ES5.

## Fixes

#604 

## Checklist

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es5",
+        "target": "ES2019",
         "strict": true,
         "downlevelIteration": true,
         "noImplicitReturns": true,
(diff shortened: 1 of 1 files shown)


mildsunrise:
lightly tested on latest 10.x and it seems to work correctly, although there are some unrelated crashes. do we officially support 10.x or is it not worth to debug?

thegecko:
I'm keen we retain backwards compatibility as much as possible. I think the node version is somewhat dictated by the napi support

thegecko:
Happy with this change in theory. I'll do some hardware tests when I get the opportunity.

Issue #603, opened 2023-04-29
switch out eventemitter for eventtarget


jimmywarting:
you know... in order to follow the spec more...

mildsunrise:
dropping support for Node.js < 14.5.x just for a detail in the hotplug detection API seems not worth it for me... i'd give more priority to stuff like switching Error for DOMError imho

jimmywarting:
node v20 just got released so support for v14 will end in 2023-04-30
v16 is the LTS now.

mildsunrise:
yeah but this is a big project and y'know. not sure what we're really supporting in practice, but at least in theory we're adamant to drop support for old Nodes. not saying I wouldn't consider it though. @thegecko thoughts?

thegecko:
I agree @mildsunrise. While I'm not adamant we support older node versions forever, I'm also keen we break people as little as possible.

Are there any other benefits we can get through dropping older node support which would make this worthwhile?

thegecko:
I'm wondering if we can at least drop node 10 support and allow to N-API 8 features

PR #606, opened 2023-05-01
Drop node 10 support and support NAPI 8

thegecko:
## Changes

With reference to discussion in #603 and #605, this PR removes support for node 10 and allows NAPI 8 features.

## Checklist

Have you...

Hardware tests run on node (macOS):

- 12.22.7
- 14.18.0
- 16.14.2

cc @mildsunrise 
Would we want a major or minor version bump on release?

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -34,7 +34,7 @@
       'defines': [
         '_FILE_OFFSET_BITS=64',
         '_LARGEFILE_SOURCE',
-        'NAPI_VERSION=6'
+        'NAPI_VERSION=8'
       ],
       'include_dirs+': [
         'src/',
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "version": "2.9.0",
   "main": "dist/index.js",
   "engines": {
-    "node": ">=10.20.0 <11.x || >=12.17.0 <13.0 || >=14.0.0"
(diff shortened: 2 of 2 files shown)


mildsunrise:
hmm, is there a reason in particular for NAPI 8? IIRC, we required NAPI 6 for context awareness (in particular for the env data)

thegecko:
No reason, I just used the napi 8 node requirements to inform what is supported.

This change allows napi 8 features if we want to use them in future.

mildsunrise:
I see... for reference, the only N-API >6 feature that IMHO we could need in the future is async cleanup hooks, but I don't think it'll come to that. still I'd like to drop support someday to force people to upgrade, maybe with some notice?

thegecko:
> I'm not sure I follow

Sorry, I'll try to explain...

Dropping node 10 gives us the engine range:

```
>=12.17.0 <13.0 || >=14.0.0
```

However, if we tweak this slightly to:

```
>=12.22.0 <13.0 || >=14.17.0
```

It means we can allow N-API v8 features in future without worry of it breaking.

I'm happy to just drop node 10 for now and we can change the N-API version when we need to instead. WDYT?

mildsunrise:
oh I see... if we drop support for Node 10, I think it's reasonable to also require 12.22 instead of 12.17 and this way we get the N-API features too. it's only a few minor versions :)

PR #611, opened 2023-05-04
Add reference to libusb FAQ

### LINKED ISSUE #182: Add Troubleshooting FAQ
### Runtime errors
* LIBUSB_ERROR_BUSY
  * Linux: detach_kernel_driver (Example in #174)
* LIBUSB_ERROR_ACCESS
  * macOS: Can't open devices that have a kernel driver, no good workaround. (discussion on #30)
     * Usually comes up with HID devices -- is there a recommended libhid library?
  * Linux: permissions, udev rules
* LIBUSB_ERROR_NOT_SUPPORTED
  * Windows: Bind WinUSB driver with libwdi / Zadig
    * Mention WCID for custom devices
* LIBUSB_ERROR_NO_DEVICE
   * macOS doesn't configure devices with bDeviceClass=255 "vendor defined" (#61)
     * If it's your firmware, use bDeviceClass=0 "defined at interface level" instead
     * workaround is `.open(false)`, `.setConfiguration(1)` to tell the kernel to configure the device.

### Install errors
* Install fails with `gyp WARN EACCES user "root" does not have permission to access the dev dir` (#175, #210, #218)
   * Don't run npm as root. Best way is to install node via nvm, which sets up permissions correctly to run `npm -g` as your normal user.
   * Or if you really have to, `npm install --unsafe-perm usb`.
   * Common issue for all NPM packages with native code nodejs/node-gyp#454 (comment)

thegecko:
## Changes

- Add reference to libusb FAQ, covering the majority of troubleshooting issues

## Fixes

Fixes #182 

## Checklist

Have you...

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ You can install one using [Zadig](http://zadig.akeo.ie/) or another approach is 
 
 Note that you cannot use multiple drivers on Windows as they get exclusive access to the device. So if you want to switch between drivers (e.g. using a printer with this software or the system), you will need to uninstall/install drivers as required.
 
+For further info, check [How to use libusb on Windows](https://github.com/libusb/libusb/wiki/Windows#user-content-How_to_use_libusb_on_Windows) in the libusb's wiki.
+
 ## Linux
 
 On Linux, you'll need libudev to build libusb if a prebuild is not available. On Ubuntu/Debian:
@@ -27,6 +29,10 @@ On Linux, you'll need libudev to build libusb if a prebuild is not available. On
 ```bash
 sudo apt-get install build-essential libudev-dev
 ```
+
+# Troubleshooting
+
+For libusb issues, please refer to the FAQ at https://github.com/libusb/libusb/wiki/FAQ
 
 # Installation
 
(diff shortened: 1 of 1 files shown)


mildsunrise (approved):
Just a suggestion, but maybe we could move that link up into the "prerrequisites" section? I say this because we already say some troubleshooting-like stuff there, such as

> if you get LIBUSB_ERROR_NOT_SUPPORTED when attempting to open your device, it's possible your device doesn't have a WinUSB driver for libusb to use

so it may be a better / more visible place. Also now that I think of it, at the end of the "Windows" subsection maybe we could also say something like:

> For further info, check [How to use libusb on Windows](https://github.com/libusb/libusb/wiki/Windows#user-content-How_to_use_libusb_on_Windows) in the libusb's wiki.

mcuee:
If you have some comments or improvement ideas for libusb FAQ, you are welcome to comment here.
* https://github.com/libusb/libusb/issues/1253

thegecko:
Is that better @mildsunrise ?

mildsunrise (approved):
yes, thank you :)

PR #631, opened 2023-06-07
Fix startPolling to use callback vs over emitted errors.

### LINKED ISSUE #630: startPoll should utilize callback for errors vs emitting them
## Bug Description:
Similar to #548 there are times (many on Windows) where a hot unplug event can cause multiple errors to become emitted. Originally in my fix for #548, I never ran into this with the prototype devices I had, however now during hot unplugs when a transfer is actively occurring, many errors tend to be emitted.

These should occur as a callback, which is currently not used but exists in the code. This allows the calling process to ensure they catch all errors, and moves away from a pattern which was causing `FATAL ERROR` to be thrown.

## Steps to Reproduce:
1. It's mildly difficult, but the same code from #548 can cause this, assuming you have a usb device that is talkative and doesn't shutdown nicely (e.g. -- mid transfer)
2. Shut down device, repeat, often the first or second time will cause multiple errors to emit
3.

## Additional Information
- Operating System: This occurred for me on Windows 11 and Ubuntu 22, under the electron environment and normal node env.
- Node USB Version (`npm ls usb`): usb@2.9.0

strazzere:
## Changes
- Use callback in `startPoll` which is not being used vs emitting errors blindly

## Fixes
- Fixed #630 by utilizing the callback which is not being used vs emitting errors

## Checklist

The only thing this impacts is most people likely never used the callback, as it was never utilized, however it seems like it was meant to be. This means people (like myself) likely are adding an error catching for the emits. Though this is what causes the issues with FATAL ERRORS.

### DIFF
--- a/test/usb.coffee
+++ b/test/usb.coffee
@@ -187,8 +187,7 @@ describe 'Device', ->
                     throw e
 
                 inEndpoint.on 'end', ->
-                    #console.log("Stream stopped")
-                    done()
+                    assert.equal(n, 100)
 
         describe 'OUT endpoint', ->
             outEndpoint = null
--- a/test/usb.coffee
+++ b/test/usb.coffee
@@ -187,7 +187,7 @@ describe 'Device', ->
                     throw e
 
                 inEndpoint.on 'end', ->
-                    assert.equal(n, 100)
+                    assert.equal(pkts, 100)
 
         describe 'OUT endpoint', ->
             outEndpoint = null
--- a/test/usb.coffee
+++ b/test/usb.coffee
@@ -172,10 +172,12 @@ describe 'Device', ->
             it 'polls the device', (done) ->
                 pkts = 0
 
-                inEndpoint.startPoll 8, 64, (e, b, n) ->
+                inEndpoint.startPoll 8, 64, (e, b, a, c) ->
+                    assert.equal(c, true)
                     assert.ok(e == undefined, e)
-                    assert.equal(n, 100)
+                    assert.equal(pkts, 100)
                     done()
+
                 inEndpoint.on 'data', (d) ->
                     assert.equal d.length, 64
                     pkts++
--- a/test/usb.coffee
+++ b/test/usb.coffee
@@ -172,7 +172,10 @@ describe 'Device', ->
             it 'polls the device', (done) ->
                 pkts = 0
 
-                inEndpoint.startPoll 8, 64
+                inEndpoint.startPoll 8, 64, (e, b, n) ->
+                    assert.ok(e == undefined, e)
+                    assert.equal(n, 100)
+                    done()
                 inEndpoint.on 'data', (d) ->
                     assert.equal d.length, 64
                     pkts++
--- a/tsc/usb/endpoint.ts
+++ b/tsc/usb/endpoint.ts
@@ -119,7 +119,7 @@ export class InEndpoint extends Endpoint {
                     this.pollActive = false;
                     this.emit('end');
                     if (callback) {
-                        const cancelled = error?.errno !== LIBUSB_TRANSFER_CANCELLED;
+                        const cancelled = error?.errno === LIBUSB_TRANSFER_CANCELLED;
                         callback(cancelled ? undefined : error, buffer, actualLength, cancelled);
                     }
                 }
--- a/tsc/usb/endpoint.ts
+++ b/tsc/usb/endpoint.ts
@@ -107,7 +107,7 @@ export class InEndpoint extends Endpoint {
                     this.stopPoll();
                 }
                 if (callback) {
-                    callback(error, buffer, actualLength)
+                    callback(error, buffer, actualLength);
                 }
             }
 
--- a/tsc/usb/endpoint.ts
+++ b/tsc/usb/endpoint.ts
@@ -104,7 +104,7 @@ export class InEndpoint extends Endpoint {
                 this.emit('data', buffer.slice(0, actualLength));
             } else if (error.errno != LIBUSB_TRANSFER_CANCELLED) {
                 if (this.pollActive) {
-                    this.emit('error', error)
+                    this.emit('error', error);
                     this.stopPoll();
                 }
             }
--- a/tsc/usb/endpoint.ts
+++ b/tsc/usb/endpoint.ts
@@ -98,11 +98,11 @@ export class InEndpoint extends Endpoint {
      * @param transferSize
      * @param callback
      */
-    public startPoll(nTransfers?: number, transferSize?: number, callback?: (error: LibUSBException | undefined, buffer: Buffer, actualLength: number) => void): Transfer[] {
+    public startPoll(nTransfers?: number, transferSize?: number, callback?: (error: LibUSBException | undefined, buffer: Buffer, actualLength: number, cancelled: boolean) => void): Transfer[] {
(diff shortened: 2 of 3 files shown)


strazzere:
I'll see what I can do around simplifying and getting a unit test out for
this. I may have some time later today to dig into it.

However I think this is primarily just an architectural issue. When a
callback is used, when available, there is no chance to raise a (fatal)
exception.

However when the errors are emitted, there is a chance that there listened
are already removed, as error and end (in that order) can already fired
have fired. One would assume after catching an error and then end, that it
is safe to remove the handlers as the event processing should be done, but
it is not guaranteed.

Utilizing the callback seemed like the most effective way to walk this back
into a better pattern.

strazzere:
Add a slim test to ensure the callback is done and called correctly -- also ordered a bluepill device to see if I can trigger the fault condition and ensure we are handling it correctly. This hopefully will get delivered within a week or so.

test/usb.coffee
```diff
@@ -172,7 +172,10 @@ describe 'Device', ->
             it 'polls the device', (done) ->
                 pkts = 0
 
-                inEndpoint.startPoll 8, 64
+                inEndpoint.startPoll 8, 64, (e, b, n) ->
+                    assert.ok(e == undefined, e)
```
thegecko, line 176: I don't think this will get called unless there is an error (in which case the assertion will fail!)

Should we be calling the callback after startPoll succeeds?

thegecko (changes requested):
I think we may need to call the callback when polling is done

thegecko:
> ...ordered a bluepill device to see if I can trigger the fault condition and ensure we are handling it correctly.

Awesome, thanks!

I can confirm the tests all still pass, but the new test also passes without your changes :)

test/usb.coffee
strazzere: That's a fair point, I was hyper fixated on my case where it was due to the error being emitted always, but yes, this should just call the callback during normal and errors cases, allowing the user to handle it.

strazzere: I believe the assertion that there is currently no error (which the test expects) is valid, but I fixed it so it will always call the callback upon emitting an end, with any error it may have as well.

tsc/usb/endpoint.ts
```diff
@@ -117,6 +118,9 @@ export class InEndpoint extends Endpoint {
                     this.pollTransfers = [];
                     this.pollActive = false;
                     this.emit('end');
+                    if (callback) {
+                        callback(error, buffer, actualLength);
```
thegecko, line 122: I wonder if we should check for cancelled here and set error to be undefined if it was cancelled.

e.g.
```typescript
if (error.errno === LIBUSB_TRANSFER_CANCELLED) {
    error = undefined;
}
```

tsc/usb/endpoint.ts
```diff
@@ -96,14 +96,15 @@ export class InEndpoint extends Endpoint {
...
      */
-    public startPoll(nTransfers?: number, transferSize?: number, _callback?: (error: LibUSBException | undefined, buffer: Buffer, actualLength: number) => void): Transfer[] {
+    public startPoll(nTransfers?: number, transferSize?: number, callback?: (error: LibUSBException | undefined, buffer: Buffer, actualLength: number) => void): Transfer[] {
         const transferDone = (error: LibUSBException | undefined, transfer: Transfer, buffer: Buffer, actualLength: number) => {
             if (!error) {
                 this.emit('data', buffer.slice(0, actualLength));
             } else if (error.errno != LIBUSB_TRANSFER_CANCELLED) {
-                this.emit('error', error);
                 if (this.pollActive) {
+                    this.emit('error', error)
```
thegecko, line 107:
```suggestion
                    this.emit('error', error);
```

test/usb.coffee
```diff
@@ -184,8 +187,7 @@ describe 'Device', ->
                     throw e
 
                 inEndpoint.on 'end', ->
-                    #console.log("Stream stopped")
-                    done()
+                    assert.equal(n, 100)
```
thegecko, line 190: n doesn't exist here

thegecko (changes requested):

tsc/usb/endpoint.ts
strazzere: I'm not positive, however at least in my current workflow, it actually helps to know that it was explicitly cancelled vs some other type of error (for ignoring purposes downstream)

thegecko: Hmm, I see your point. If we keep it like that, we'd need to document that cancelling a poll throws an error :/

strazzere: So, to be very clear. It doesn't `emit` nor does it `throw` an error when a `LIBUSB_TRANSFER_CANCELLED` is encountered. With the above changes though, it would now pass it back in the call back upon the finish of the `startPoll`.

This is good, as in prior to this code, one would never know this error even occurred at all.

thegecko: It is common practice to check for an error in callback and assume failure, though. If the function was [promisified](https://nodejs.org/api/util.html#utilpromisifyoriginal) I believe it would result in an error being thrown.

strazzere: Agreed. It is common, but not necessarily the only way to handle the situation depending on the error. Error's are meant to be fatal/unrecoverable in those situations. As far as I can tell the `LIBUSB_TRANSFER_CANCELLED` isn't necessarily and `error` in the sense that we are talking about at all.

This is all my understanding of the libusb stack and how it's being utilized here, so I could be way off map here.

I still think you're correct in stating we need to document it, I just wanted to be clear it isn't a literal `throw` or `emit`. Personally I'd argue that this should actually be a `boolean` getting sent back? Though this is also a large deviation from the current usage pattern.

thegecko: 👍 I like this proposal.  Although it changes the signature, adding an extra parameter doesn't break backwards compatibility.

I assume in the case of `LIBUSB_TRANSFER_CANCELLED`,  the error would be undefined and we'd set the cancelled boolean?

strazzere: Got it, I can make these changes likely later this week. Bluepill arrived as well but I haven't had any time to toy around with it or the testing firmware repo.

strazzere: I've pushed code for this, obviously adding a test to the firmware would be best. Though this will take more cycles, which I'm planning, but don't have at the moment. Feel free to just ignore the PR until then if you'd like. I can always add test cases via another PR later as well.

test/usb.coffee
strazzere: Fixed in 8487be6

strazzere:
This should be ready for (hopefully) final review.

strazzere:
@thegecko No worries if not, and sorry if I'm pestering, however I'm wondering if you had a chance to skim over this?

thegecko:
Apologies for the delay on this, I've been away.

I'll take a look at the PR and tests when I find some time...

strazzere:
No worries, was just trying to bump it before I forgot everything I did for it!

thegecko:
Thanks, I've confirmed your tests have added the expected behaviour.

I've also re-enabled the test for using events with polling so both forms are checked:

https://github.com/node-usb/node-usb/commit/d3663ad1afa8a4d9175d8da442f4f35ab31915c5

PR #726, opened 2024-01-17
fix: uncaught error when loading in wsl #724

### LINKED ISSUE #724: WSL "Uncaught TypeError"
## Bug Description:
It is already known that this library does not function in WSL, due to libusb failing to initialise (not the fault of this library).

When this happens it logs `Failed to initialize libusb.`, and then it attempts to use `usb.Device` and throws an error.
```
Uncaught TypeError: Cannot read properties of undefined (reading 'prototype')
    at /home/julus/companion/node_modules/usb/dist/usb/index.js:19:38
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (/home/julus/companion/node_modules/usb/dist/usb/index.js:18:63)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:121:18)
>
```

This thrown error makes it hard to gracefully handle this, as the whole require/import needs to be wrapped in a try/catch, rather than checking if the library looks sane at the point where it is used

## Steps to Reproduce:
1. Launch WSL and install nodejs
2. Create a new project and add `usb` as a dependency
3. run a script which imports the `usb` library and observe the error

## Additional Information
- Operating System: Ubuntu 20.04 in WSL
- Runtime: 18.19.0
- Node USB Version (`npm ls usb`): 2.10.0

Julusian:
## Changes

- Fix #724

## Fixes

- #724

## Checklist

Have you...

### DIFF
--- a/tsc/usb/index.ts
+++ b/tsc/usb/index.ts
@@ -18,9 +18,12 @@ Object.defineProperty(usb, 'pollHotplugDelay', {
     writable: true
 });
 
-Object.getOwnPropertyNames(ExtendedDevice.prototype).forEach(name => {
-    Object.defineProperty(usb.Device.prototype, name, Object.getOwnPropertyDescriptor(ExtendedDevice.prototype, name) || Object.create(null));
-});
+// `usb.Device` is not defined when `usb.INIT_ERROR` is true
+if (usb.Device) {
+    Object.getOwnPropertyNames(ExtendedDevice.prototype).forEach(name => {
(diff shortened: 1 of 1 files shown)


thegecko (approved):

thegecko:
LGTM, thanks!

Issue #61, opened 2014-12-24
LIBUSB_ERROR_NOT_FOUND on MacOS when trying to open a device.


jmbldwn:
Kinda takes the fun out of it.

I'm able to find devices just fine:

> var usb = require('usb');
> var dev = usb.findByIds(0x132b, 0x2101);
> dev
{ busNumber: 20,
  deviceAddress: 12,
  deviceDescriptor: 
   { bLength: 18,
     bDescriptorType: 1,
     bcdUSB: 272,
     bDeviceClass: 255,
     bDeviceSubClass: 0,
     bDeviceProtocol: 0,
     bMaxPacketSize0: 16,
     idVendor: 4907,
     idProduct: 8449,
     bcdDevice: 256,
     iManufacturer: 1,
     iProduct: 2,
     iSerialNumber: 0,
     bNumConfigurations: 1 } }

... but trying to open the device fails:

> dev.open();
Error: LIBUSB_ERROR_NOT_FOUND
    at Device.Object.defineProperty.get (/Users/jim/node_modules/usb/usb.js:45:45)
    at Device.usb.Device.open (/Users/jim/node_modules/usb/usb.js:30:16)
    at repl:1:5
    at REPLServer.self.eval (repl.js:110:21)
    at repl.js:249:20
    at REPLServer.self.eval (repl.js:122:7)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.emit (events.js:95:17)
    at Interface._onLine (readline.js:202:10)
    at Interface._line (readline.js:531:8)

libusb1.0.19 installed.  Also tried libusb1.0.8.  Same answer.

Ideas?

jtbonhomme:
Same issue for me:
OSX Lion 10.7.5
```
$  uname -a
Darwin new-host.home 11.4.2 Darwin Kernel Version 11.4.2: Thu Aug 23 16:25:48 PDT 2012; root:xnu-1699.32.7~1/RELEASE_X86_64 x86_64
$  node -v
v0.10.33
$  npm -v
1.4.28
```

jmbldwn:
Here’s my workaround.  I can’t remember how I stumbled across this, but it seems to work for me.  Calling the semi-private __open() and __claimInterface() functions before the public open function seems to jiggle it into a position where it works.  Go figure.

this.open = function(callback) {
console.log('opening device')
_device = usb.findByIds(0x132b, 0x2101);

_device.__open();
_device.__claimInterface(0);

_device.open();


[...]

PR #732, opened 2024-02-10
Patches WebUSBDevice to force reconfiguration deviceClass 0xff

jounii:
## Changes

Tries to detect if deviceClass 0xff devices have not been configured by default (macOS) and then uses the workaround to set configuration manually.

Tested with Sony MDS-JB980 hardware through netmd-js library using node-usb WebUSB on macOS.

## Fixes
- #61 - So that it works for WebUSB without hacks prior using WebUSB

## Checklist

- [X] Tested the change acts as expected

### DIFF
--- a/tsc/webusb/webusb-device.ts
+++ b/tsc/webusb/webusb-device.ts
@@ -353,6 +353,11 @@ export class WebUSBDevice implements USBDevice {
             if (!this.opened) {
                 this.device.open();
             }
+            if (this.opened && this.deviceClass === 0xff && this.device.interfaces && !this.device.interfaces.length) {
+                this.device.close();
+                this.device.open(false);
+                this.device.setConfiguration(1);
+            }
 
             this.manufacturerName = await this.getStringDescriptor(this.device.deviceDescriptor.iManufacturer);
             this.productName = await this.getStringDescriptor(this.device.deviceDescriptor.iProduct);


thegecko:
Sorry this has been a bit quiet, I'm trying to decide what to do.

Concerns:
- The forced reopen doesn't feel right
- However, neither does exposing the `open(false)` functionality through webusb
- you can't check for interfaces without it being opened initially
- `setConfiguration` is already exposed on the webusb class

Would this be a better approach?

```typescript
            if (!this.opened) {
                // Don't read config descriptor of vendor-specific devices on macos
                // https://github.com/node-usb/node-usb/issues/61
                if (this.deviceClass === 0xff && platform() === 'darwin') {
                    this.device.open(false);
                } else {
                    this.device.open();
                }
            }
```

And then call `selectConfiguration` on webusb later? Or does that still fall over due to needing `device.configDescriptor` @jounii ?

jounii:
> Concerns:
> 
> * The forced reopen doesn't feel right
> * However, neither does exposing the `open(false)` functionality through webusb

Yes I wasn't happy about it either, but couldn't figure any other way besides having non-standard callback exposed. A bit like the `devicesFound`, but for `beforeOpen` or `afterOpen` or maybe `configureDevice` for doing manually the re-open.. Or maybe just a flag to allow "forced reopen" or so. The `platform()` check sounds good to limit to macOS. The existing `devicesFound` comes too late as it expects the device being opened successfully and listed.

> * you can't check for interfaces without it being opened initially
> * `setConfiguration` is already exposed on the webusb class
> 
> And then call `selectConfiguration` on webusb later? Or does that still fall over due to needing `device.configDescriptor` @jounii ?

I will test, but I recall it did not find the device unless the interfaces array was initilized and without the setConfiguration the device was basically considered non-existing as interfaces was empty array and thus considered "opened", yet not really opened.

thegecko:
I've reimplemented your fix in https://github.com/node-usb/node-usb/pull/739 , which i think is a little neater.

I'm not able to test this, but can't see why it wouldn't work and fix the issue. Feedback is helpful though!

jounii:
> I've reimplemented your fix in #739 , which i think is a little neater.
> 
> I'm not able to test this, but can't see why it wouldn't work and fix the issue. Feedback is helpful though!

Looks great and neat solution. I will test. I have been occupied with other things and didn't yet get back to this.

thegecko:
Closing this in favour of #739

PR #765, opened 2024-05-02
Upgrade prebuildify to 6.0.1

thegecko:
## Changes

-

## Fixes

-

## Checklist

Have you...


strazzere:
Fwiw, this will still suffer not properly identifying prebuilts until the original bug that we rolled back for. Per my comment to https://github.com/node-usb/node-usb/pull/764#issuecomment-2091498614 - I don't believe bumping `node-gyp` fixes it still.

PR #809, opened 2024-09-09
Update device.cc

cleoo:
Avoid node js fatal exception

## Changes
Send exception back to node

## Fixes
fix "Abandon (core dumped)" in certain conditions, including unexpected disappearances of devices from the usb bus. The errors are traced back to node, which makes these errors catchable

## Checklist

### DIFF
--- a/src/device.cc
+++ b/src/device.cc
@@ -246,7 +246,8 @@ struct Req: Napi::AsyncWorker {
             Callback().Call(device->Value(), { error });
         }
         catch (const Napi::Error& e) {
-            Napi::Error::Fatal("", e.what());
+            // Napi::Error::Fatal("", e.what());
+            THROW_ERROR(e.what());
         }
     }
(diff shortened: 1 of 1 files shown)


cleoo:
The error messages returned have little to do with my pull request.

cleoo:
Hi, could I have an idea of ​​the integration time of these changes? I just tested this morning on a sample of machines and all my errors now go through process.on('uncaugthException'), it is an open door to correct processing of these exceptions that were previously invisible. I ask the question of the time because I have about 800 machines to update

thegecko:
Merges are blocked on this being fixed:

https://github.com/prebuild/prebuildify-cross/issues/20

We need prebuilds to pass before merging and releasing.

thegecko:
@cleoo can you merge your PRs into one, please and add a descriptive title.

Issue #672, opened 2023-09-16
Uncatchable LIBUSB_TRANSFER_NO_DEVICE


cleoo:
## Bug Description:
We use node-usb to drive escpos thermal printers. It sometimes happens that certain models that are faulty or come with faulty firmware disconnect randomly from the usb bus.

When the disconnection occurs, the exception generated is impossible to catch and stops nodejs, which is not really acceptable.

## Error logged

> FATAL ERROR:  LIBUSB_TRANSFER_NO_DEVICE
 1: 0xb02960 node::Abort() [node /home/kiosk/[A]/index.js]
 2: 0xa18149 node::FatalError(char const*, char const*) [node /home/kiosk/[A]/index.js]
 3: 0xa18152  [node /home/kiosk/[A]/index.js]
 4: 0xaccedb napi_fatal_error [node /home/kiosk/[A]/index.js]
 5: 0x7f9da5e15046  [/home/kiosk/[B]/node_modules/usb/prebuilds/linux-x64/node.napi.glibc.node]
 6: 0x7f9da5e2440b handleCompletion(Transfer*) [/home/kiosk/[B]/node_modules/usb/prebuilds/linux-x64/node.napi.glibc.node]
 7: 0x7f9da5e154bd Napi::ThreadSafeFunction::CallJS(napi_env__*, napi_value__*, void*, void*) [/home/kiosk/[B]/node_modules/usb/prebuilds/linux-x64/node.napi.glibc.node]
 8: 0xaca7e0  [node /home/kiosk/[A]/index.js]
 9: 0x1547da6  [node /home/kiosk/[A]/index.js]
10: 0x155a4d4  [node /home/kiosk/[A]/index.js]
11: 0x15486f8 uv_run [node /home/kiosk/[A]/index.js]
12: 0xa3fa35 node::SpinEventLoop(node::Environment*) [node /home/kiosk/[A]/index.js]
13: 0xb42c56 node::NodeMainInstance::Run() [node /home/kiosk/[A]/index.js]
14: 0xac89bc node::Start(int, char**) [node /home/kiosk/[A]/index.js]
15: 0x7f9dc8227cd0  [/usr/lib/libc.so.6]
16: 0x7f9dc8227d8a __libc_start_main [/usr/lib/libc.so.6]
17: 0xa3d03c  [node /home/kiosk/[A]/index.js]
## Steps to Reproduce:
1. Possibly disconnect a working device during an info transfer

## 
## Additional Information
[...]

PR #817, opened 2024-09-13
fix fatal exception not thrown in javascript context

cleoo:
## Changes

- Stop using Napi::Error::Fatal exceptions which seems to break the vm

## Fixes

- https://github.com/node-usb/node-usb/issues/672

## Checklist

Have you...

### DIFF
--- a/src/device.cc
+++ b/src/device.cc
@@ -246,7 +246,7 @@ struct Req: Napi::AsyncWorker {
             Callback().Call(device->Value(), { error });
         }
         catch (const Napi::Error& e) {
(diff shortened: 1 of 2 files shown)


thegecko:
Thanks, I can confirm this passes all the tests. Can you think of any way to add a test to show what this fixes (and avoid future regressions)?

cleoo:
Hi,

The “non-catchable” error was originally triggered by the unexpected disappearance of a device from the bus. Perhaps the test could be simulated by executing a “usbreset XXX/XXX” command? I'm having a bit of trouble seeing how to set up a unit test in this scenario.

Issue #803, opened 2024-08-28
getDeviceList() does not update the list of devices properly


JonathanLeeIFX:
## Bug Description:
getDeviceList() will not show a device's new address if it is connected, removed, then reconnected.

## Steps to Reproduce:
I tested this using "usbview.exe" to check the Device Address and a Typescript program that tries to find a Silicon Labs C8051 device with  VID=0x10C4 and PID=0xEA61 every 3 seconds.
1. Connect the device and check the Device Address in "usbview.exe".
2. Launch the Typescript program. Observe that it can detect the device and reports the same Device Address.
3. Disconnect the device. Wait 2 seconds.
4. Reconnect the device. Observe that "usbview.exe" shows a different Device Address for the C8051, but the Typescript program continues to report the old Device Address.

## Additional Information
- Operating System: Windows 10
- Runtime: Node v16.17.0
- Node USB Version: usb@2.13.0

## Typescript to reproduce:

```
import * as usb from 'usb';
async function main() {
	let i = 0;
	do {
		await delay(3000);	
		const devices = usb.findByIds(<VENDOR_ID>, <PRODUCT_ID>);		
		console.log(`device run ${i}: `, devices);
		i++;
	} while (true);
}
function delay(ms: number) {
    return new Promise( resolve => setTimeout(resolve, ms) );
}
main();
```

thegecko:
Thanks for raising this and the reproducer code
I can confirm I see the same behaviour and the issue is apparent on windows 10. It looks like a race condition.
It doesn't seem to occur on macos, so seems seems to be OS specific.

I've had a play around with the code and even invalidating the internal cache doesn't fix the issue.

Breaking it down to the fewest calls with [libusb_get_device_list](https://github.com/node-usb/node-usb/blob/main/src/node_usb.cc#L104) and printing the address returned with [libusb_get_device_address](https://github.com/node-usb/node-usb/pull/821/files#diff-ba21e282fe0588b3a65e3ab4a5b27a48a13c68f2d4dfa7958567b90a4f96f5d1R110) still shows no change between disconnect/connect.

This leads me to believe its an issue with the `libusb` library on windows and needs to be logged [there](https://github.com/libusb/libusb/issues).

JonathanLeeIFX:
@thegecko After some discussion with **tormodvolden**, I've created an example project using C and the original libusb.dll: [libusbTest](https://github.com/JonathanLeeIFX/libusbTest).

[...]

Issue #579, opened 2023-03-09
getDevices() list incomplete


axelzuzek:
## Bug Description:

 webusb.getDevices(); doesn't find all connected usb devices

## Steps to Reproduce:
```
   const usb = require('usb');

   const webusb = new WebUSB({
        allowAllDevices: true
    });

    const showDevices = async () => {
        const devices = await webusb.getDevices();
        const text = devices.map(d => `${d.productName}\t${d.productId}`);
        text.unshift('VID\tPID\tSerial\n-------------------------------------');
        console.log(text.join('\n'));
    };

    showDevices();
``
lists for example my Razer Lancehead mouse, my keyboard, but not my Razer Deathadder V2 Pro mouse.

<!-- Please provide any additional information available. -->
<!-- Additional information can be in the form of logs, screenshots, screencasts. -->
<!-- If possible, help to bisect the issue by outlining versions the issue didn't occur in. -->
## Additional Information
electron 23.1.3
- Operating System: Windows 10
- Runtime node v16.13.1
- Node USB Version  usb@2.6.0

**Comment by thegecko** [comment]
Mice and keyboards are HID devices so are usually grabbed by the operating systems and aren't available through LibUSB. To access those devices I recommend you look at [node-hid](https://github.com/node-hid/node-hid).

There are also cases where LibUSB devices can't be turned into WebUSB devices so rather than throw an error while listing, the errors are [silently handled](https://github.com/node-usb/node-usb/blob/master/tsc/webusb/index.ts#L269).

To test your case, you can list the native devices LibUSB has access to:
https://github.com/node-usb/node-usb#list-all-legacy-devices

And if the device you need is there, you can see any error thrown when turning it into a WebUSB device by using:

```typescript
import { getDeviceList, WebUSBDevice } from 'usb';

(async () => {
    const devices = getDeviceList();
    const device = devices[0]; // Or find your device using serial, vid/pid, etc.

    // Uses blocking calls, so is async
    const webDevice = await WebUSBDevice.createInstance(device); // This will thow any creation errors

    if (webDevice) {
        console.log(webDevice); // WebUSB device
    }
})();
```

PR #828, opened 2024-10-15
getDeviceList() returns stale devices

YishaiYosifov:
## Changes

Compiling libusb without udev fixes the problem.
I tried running [libusbTest](https://github.com/JonathanLeeIFX/libusbTest) from https://github.com/node-usb/node-usb/issues/803#issuecomment-2389707705 and it didn't work either - unless I compiled libusb without udev.

## Fixes
- #579 

## Checklist

Have you...

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,6 +1,6 @@
 {
   'variables': {
-    'use_udev%': 1,
+    'use_udev%': 0,
     'use_system_libusb%': 'false'
   },
   'dependencies': [
--- a/libusb.gypi
+++ b/libusb.gypi
@@ -1,6 +1,6 @@
 {
   'variables': {
-    'use_udev%': 1,
+    'use_udev%': 0,
     'use_system_libusb%': 'false',
     'module_name': 'usb_bindings',
     'module_path': './src/binding'


thegecko:
Thats odd, any idea why this is the case?

YishaiYosifov:
(●__● ) great question, I have no idea
I found similar problems here: https://libusb-devel.narkive.com/pqWf4Hig/libusb-libusb-get-device-list-is-stale-and-does-not-refresh and here: https://libusb-devel.narkive.com/pqWf4Hig/libusb-libusb-get-device-list-is-stale-and-does-not-refresh

YishaiYosifov:
@thegecko oops I accidently attached the wrong issue - I meant #803 haha

thegecko:
I think (please correct me if this is wrong):

- udev is only used on linux (maybe mac, but certainly not windows)
- the issue we are seeing is only on windows

Therefore, the correct fix may be to disable udev for windows builds only.

WDYT?

YishaiYosifov:
I was running it on Linux...

thegecko:
> I was running it on Linux...

Oh, the original issue was on windows

thegecko:
@JonathanLeeIFX can you confirm if this fixes your issue on Windows?

JonathanLeeIFX:
It's interesting that the example program did not work for you. It's supposed to be working because it only uses libusb.

This fix did not work for me, but I may have done it wrong.
My procedure:
1. Navigate to my node-usb test project directory.
2. Open 'node_modules/usb/binding.gyp', change line 3 to "'use_udev%': 0,".
3. Open 'node_modules/usb/libusb.gypi', change line 3 to "'use_udev%': 0,".
4. Recompile project with 'tsc index.ts'
5. Run with 'node index.js'.

Output (edited for readability)
**device run 1**:  {
 ...
  deviceAddress: 19,
 ...
}

**device run 2**:  {
...
  deviceAddress: 19,
...
}
**device run 3**:  undefined
**device run 4**:  undefined
**device run 5**:  {
...
  deviceAddress: 19,
...
**device run 6**:  {
...
  deviceAddress: 19,
...
}

YishaiYosifov:
Did you recompile the binaries as well? Try the `prebuild` script, that's what I was using when testing it

JonathanLeeIFX:
navigating to 'node_modules\usb' and running 'npm run prebuild' gives me the following error:

> usb@2.13.0 prebuild
> prebuildify --napi --strip --name node.napi

gyp info it worked if it ends with ok
gyp info using node-gyp@10.2.0
gyp info using node@20.15.0 | win32 | x64
gyp info find Python using Python version 3.10.0 found at "C:\Users\LeeJo\AppData\Local\Programs\Python\Python310\python.exe"

gyp http GET https://nodejs.org/dist/v23.0.0/node-v23.0.0-headers.tar.gz
gyp http 404 https://nodejs.org/dist/v23.0.0/node-v23.0.0-headers.tar.gz
gyp WARN install got an error, rolling back install
gyp ERR! configure error
gyp ERR! stack Error: 404 response downloading https://nodejs.org/dist/v23.0.0/node-v23.0.0-headers.tar.gz
gyp ERR! stack at go (C:\Users\LeeJo\Documents\Infineon\test-repository\node_modules\usb\node_modules\node-gyp\lib\install.js:223:21)
gyp ERR! stack at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
gyp ERR! stack at async install (C:\Users\LeeJo\Documents\Infineon\test-repository\node_modules\usb\node_modules\node-gyp\lib\install.js:63:18)
gyp ERR! stack at async getNodeDir (C:\Users\LeeJo\Documents\Infineon\test-repository\node_modules\usb\node_modules\node-gyp\lib\configure.js:79:7)
gyp ERR! stack at async run (C:\Users\LeeJo\Documents\Infineon\test-repository\node_modules\usb\node_modules\node-gyp\bin\node-gyp.js:81:18)
gyp ERR! System Windows_NT 10.0.19045
gyp ERR! command "c:\\program files\\nodejs\\node.exe" "C:\\Users\\LeeJo\\Documents\\Infineon\\test-repository\\node_modules\\usb\\node_modules\\node-gyp\\bin\\node-gyp.js" "rebuild" "--targ
et=23.0.0" "--devdir=C:\\Users\\LeeJo\\AppData\\Local\\Temp\\prebuildify\\node" "--arch=x64" "--release"
gyp ERR! cwd C:\Users\LeeJo\Documents\Infineon\test-repository\node_modules\usb
gyp ERR! node -v v20.15.0
gyp ERR! node-gyp -v v10.2.0
gyp ERR! not ok
node-gyp exited with 1

YishaiYosifov:
hmm... Try to clone to repo, run `prebuild` and copy the `prebuilds` folder to your node_modules/usb folder

JonathanLeeIFX:
I cloned the repo and it failed on 'npm install'. I had to copy the 'libusb' folder in from my original test repo. After that, 'npm install' worked.

I then changed 'libusb.gypi' and 'binding.gyp', ran 'npm run prebuild', and copied 'prebuilds\win32-x64' to my test repo.
No change.

YishaiYosifov:
It works on my linux machine:

```ts
import { getDeviceList } from "usb";

async function main() {
    while (true) {
        const allDevices = getDeviceList();
        const device = allDevices.find(
            (device) => device.deviceDescriptor.idProduct === <PID>
        );
        console.log(device?.deviceAddress);
        await new Promise((resolve) => setTimeout(resolve, 1000));
    }
}

main();

```

image

Maybe this fix only works on linux?

JonathanLeeIFX:
Possibly, but I'm still confused about the root issue: why does the node-usb package have an issue but not the original libusb DLL?
'getDeviceList()' should be a direct call to the DLL function. Why would it matter where it's being called from?

YishaiYosifov:
Like I said in the PR, the original libusb DLL also didn't work for me

YishaiYosifov:
@thegecko I think I found the problem on windows (I don't know c++ that well, so please correct me if I get anything wrong).
In the device constructor [we are calling `libusb_ref_device`](https://github.com/node-usb/node-usb/blob/main/src/device.cc#L15), which increments the ref count. Then, when we call `libusb_free_device_list` it only decrements it by 1, meaning it is still reffed.

If I add `libusb_unref_device` [right after constructing the device](https://github.com/node-usb/node-usb/blob/main/src/node_usb.cc#L111) it works, but only until the device is garbage collected in js, `libusb_unref_device` is called again in the deconstructor and then [this assert](https://github.com/libusb/libusb/blob/master/libusb/core.c#L1296) fails.

If I remove the calls to `libusb_ref_device` in the constructor and `libusb_unref_device` in the deconstructor it works, but that's not a solution. And even after I do that is STILL doesn't work in linux unless I compile without udev

thegecko:
> In the device constructor [we are calling libusb_ref_device](https://github.com/node-usb/node-usb/blob/main/src/device.cc#L15), which increments the ref count. Then, when we call libusb_free_device_list it only decrements it by 1, meaning it is still reffed.

I looked at this too, doesn't `ref` just add 1 and `unref` remove 1 meaning net zero overall?

YishaiYosifov:
> doesn't ref just add 1 and unref remove 1 meaning net zero overall?

it starts at 1 ref

thegecko:
> it starts at 1 ref

then why does it get incremented in the constructor :/

YishaiYosifov:
I don't know man, maybe it was an oversight. If I had to guess it's because when the device is destroyed and not removed from the byPtr map it can cause some very weird behaviour in the get method.

I added a log whenever a device was destroyed in the unref method, and it only ran whenever I didn't increment it in the constructor.

thegecko:
Thanks for your help on this, I'll have a play around. This only fixes the windows issue, right? Linux is still tied up with udev.

YishaiYosifov:
> This only fixes the windows issue, right? Linux is still tied up with udev.

correct

I committed a potential fix. If the devices we actually unrefered at the end of `GetDeviceList` we won't be able to use them. so instead I temporarily unrefered them before calling `libusb_free_device_list`, then, I set the device pointer to the new pointer when getting the device. If a device still has a `nullptr` libusb device, it means the device no longer exists and we can delete it.

We also can't use the pointer as the key in the byPtr map anymore - as the pointer changes every time.
So instead I changed it to use the usb address. There is a small chance that 2 devices reconnect, and one of them gets the other ones address, but I couldn't think of a way to avoid it (as far as I can tell, there is no way to say the device is the same after reconnection).

What do you think?

thegecko:
> I changed it to use the usb address

I don't think we can rely on this, I can imagine more odd edge cases than we have at the moment :/. Perhaps we should just recreate everything as soon as a `GetDeviceList` is called?

thegecko:
> Perhaps we should just recreate everything as soon as a GetDeviceList is called?

That's not going to work is it? Any existing reads/etc. would break.

I can't help feeling we are patching over and underlying issue here. We should be able to enumerate devices again while still holding references to them. What if we create a new usb context for getting a fresh list of devices then copy the changes over?

YishaiYosifov:
> What if we create a new usb context for getting a fresh list of devices then copy the changes over?

How would we know what changed? how can we compare the devices

thegecko:
Yeah you are right.

Hmm. If we key based on `libusb_get_device_address`, I believe that's only unique per hub, so `libusb_get_bus_number` has to be in the key, too.

PR #832, opened 2024-10-31
Catch no device error when getting config descriptors

YishaiYosifov:
## Changes

When you release an interface it refreshes it inside a callback:
 image

`refresh` calls `get configDescriptor` which when the device is disconnected can throw `LIBUSB_ERROR_NO_DEVICE` .

Because this is done inside a callback I can't catch it in a try catch and it crashes the program: 
image

## Fixes

-

## Checklist

Have you...


thegecko (approved):
LGTM

thegecko:
Thanks for this fix, I'll run the hardware tests on it and merge...

PR #851, opened 2024-12-17
optional env var to override module path

barlock:
In my use of this package, I'm using `@vercel/ncc` to build my backend into a single file. Because of that, `join(__dirname, '..', '..')` is no longer accurate for where `node-gyp-build` needs to look for the binary. I have included an optional var that can be used to override the path. This seems to be standard for other binaries I'm using in my project. 

## Changes

- Introduces a new env var `NODE_USB_PATH` where one can override the expected path of the module for node-gyp-build to look in.

## Checklist

Have you...

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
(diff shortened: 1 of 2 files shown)


.idea/.gitignore
thegecko, line 1: Would you mind removing this file from your PR?

thegecko:
Thanks for this

PR #964, opened 2026-07-19
build: require C++17 for native bindings

rinchen:
## Summary

- Change `cflags_cc` from `-std=c++14` to `-std=c++17` (Linux/GCC/Clang).
- Change macOS `OTHER_CFLAGS` from `-std=c++1y` to `-std=c++17`.
- Add `/std:c++17` to Windows `AdditionalOptions` (alongside existing `/EHsc`).

## Context

Colorado-Mesh/mesh-client carries this as a pnpm patch against `usb@2.18.0` so Electron native rebuilds succeed with current toolchains that treat older dialects poorly or require C++17 APIs.

## Test plan

- [ ] Rebuild native bindings on macOS, Linux, and Windows
- [ ] Smoke-test USB device open/list after rebuild


thegecko:
Thanks for the PR @rinchen.
If we are moving to C++17, is it also worth adding flags to the Windows build?

e.g.

```
'AdditionalOptions': [ '/EHsc', '/std:c++17' ]
```

rinchen:
> Thanks for the PR @rinchen. If we are moving to C++17, is it also worth adding flags to the Windows build?

Yes. I think so, for consistency. Do you want me to add it to this PR?

thegecko:
> Yes. I think so, for consistency. Do you want me to add it to this PR?

Yes please

rinchen:
> > Yes. I think so, for consistency. Do you want me to add it to this PR?
> 
> Yes please

Done! Let me know if you want anything else. Thanks.

thegecko (approved):

